PR #1, opened 2022-02-03
Building out some of the structure for the configuration and action logic.

bsieber-mozilla:

PR #3, opened 2022-02-22
Initial Structure and Buildout of JBI

bsieber-mozilla:


leplatrem:
How does this relate to #1 ?

src/core/config.py
```diff
@@ -0,0 +1,60 @@
...
+        errors = []
+    if not config_map:
+        config_map = {}
+
+    for filename in glob.glob(os.path.join(folder_path, "*.json")):
+        try:
+            key, value = process(filename)
+            if key:
+                config_map[key] = value
+        except Exception as exception:  # pylint: disable=broad-except
```
grahamalama, line 55: Catching an `Exception` could raise an eyebrow (even the linter agrees). Are there specific errors that `process` would throw that we might want to handle differently? I do notice you have a comment at the top of the function, but maybe something closer to this line explaining why we're catching a base `Exception` would be useful.

src/core/actions.py
```diff
@@ -0,0 +1,27 @@
+from inspect import getmembers, isfunction, ismodule
+
+from src.jbi import process as jbi_process
+
+module_dict = {jbi_process.ACTION_KEY: jbi_process.actions}
+
+
+def get_action_context_by_key(key):
+    if key not in module_dict:
+        assert False, "Unknown key requested"
```
grahamalama, line 10: Do you think raising a `ValueError` here might be a better choice? From the docs:
> Raised when an operation or function receives an argument that has the right type but an inappropriate value

That description seems pretty apt for this situation.

src/core/config.py
```diff
@@ -0,0 +1,60 @@
...
+        data = json.load(file)
+        ret_dict.update(data)
+        if not ret_dict.get("enabled"):
+            # Only process enabled files
+            return None, None
+
+        action = ret_dict.get("action")
+        known_actions = actions.get_action_context_by_key(key=action_key)
+        if action not in known_actions:
+            assert False, f"CONFIG ERROR: Unknown action `{action}`."
```
grahamalama, line 22: I noticed we do this assertion with a `CONFIG ERROR` message thing a few times in this function. Maybe this would be a good usecase for making our own `ConfigError(Exception)` and raising that instead?

grahamalama (commented):
Looks like a solid start! A few inline comments, and 3 general things:

- I'm noticing that we're using assertions in places where regular exceptions might be a better choice. I'll also point out at least one or two places where I noticed it, but I think it might be a bit of a smell to use `assert` in production code. Granted, there are places where asserts can be useful, so I'm open to pushback here, but in general it's something I (personally) would avoid using

- As we kind of discussed on our call, I think some of the subpackages we're creating might not be needed just yet. Looking at it a bit more, I think all of the functions in the `core` modules could live nicely in `src.jbi.process`. I think that module has one clear job -- process the tags and actions to prepare them to be used by the FastAPI app.

- In places where you use type hints for container objects, it might be nice to further provide the type(s) for the objects in the container. E.g. `Tuple[str, str]`

src/core/config.py
```diff
@@ -0,0 +1,60 @@
...
+    # Aggregate errors; fail on blocking errors.
+    # returns dict of config
+    if not errors:
+        errors = []
+    if not config_map:
+        config_map = {}
+
+    for filename in glob.glob(os.path.join(folder_path, "*.json")):
+        try:
+            key, value = process(filename)
```
grahamalama, line 52: [pathlib](https://docs.python.org/3/library/pathlib.html) might be nice here?
```suggestion
    for filename in Path(folder_path).glob("*.json"):
        try:
            key, value = process(str(filename))
```

bsieber-mozilla:
> How does this relate to #1 ?

@leplatrem I met with graham yesterday discussing most of this PR; and will again briefly tomorrow for some general review of PR 4. 

#1 is still the primary; modifications and comments here will result in changes to #1.

This split was to reduce the amount of files viewed per-review. You can ignore #3 and #4 since it's the split history of #1 

_A future approach would be to ask for the review sooner--or split based on files/directories instead of commit history._

bsieber-mozilla:
> Looks like a solid start! A few inline comments, and 3 general things:
> 
>     * I'm noticing that we're using assertions in places where regular exceptions might be a better choice. I'll also point out at least one or two places where I noticed it, but I think it might be a bit of a smell to use `assert` in production code. Granted, there are places where asserts can be useful, so I'm open to pushback here, but in general it's something I (personally) would avoid using
> 
>     * As we kind of discussed on our call, I think some of the subpackages we're creating might not be needed just yet. Looking at it a bit more, I think all of the functions in the `core` modules could live nicely in `src.jbi.process`. I think that module has one clear job -- process the tags and actions to prepare them to be used by the FastAPI app.
> 
>     * In places where you use type hints for container objects, it might be nice to further provide the type(s) for the objects in the container. E.g. `Tuple[str, str]`

@grahamalama thanks for the review! I will respond to each comment individually (and absorb them into #1).

.gitignore
```diff
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# iSort - Pre-Commit
+.isort.cfg
```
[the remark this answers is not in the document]
bsieber-mozilla: I believe this is due to linting being used in both docker and pre-commit. IIRC precommit was the culprit that needed this.

pyproject.toml
```diff
@@ -0,0 +1,73 @@
...
+'''
+
+[tool.isort]
+# https://black.readthedocs.io/en/stable/compatible_configs.html
+multi_line_output = 3
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
+ensure_newline_before_comments = true
+line_length = 88
```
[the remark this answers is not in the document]
bsieber-mozilla: Black and iSort seem to fight via pre-commit. Whichever comes first, the other then modifies and reverses.

src/core/config.py
bsieber-mozilla: resolved in this commit:
https://github.com/mozilla/jira-bugzilla-integration/pull/1/commits/8219d4187d7cd8cde7dd014367fdf65aa6a9d98a

src/core/config.py
```diff
@@ -0,0 +1,60 @@
+import glob
+import json
+import os
+from typing import Dict, List, Tuple
+
+from src.core import actions
+
+
+def per_file_process(
+    filename: str, ret_dict: dict, required_keys, action_key, filename_key
```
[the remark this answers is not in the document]
bsieber-mozilla: resolved in this commit:
https://github.com/mozilla/jira-bugzilla-integration/pull/1/commits/8219d4187d7cd8cde7dd014367fdf65aa6a9d98a

src/core/config.py
bsieber-mozilla: resolved in this commit:
https://github.com/mozilla/jira-bugzilla-integration/pull/1/commits/8219d4187d7cd8cde7dd014367fdf65aa6a9d98a

src/core/actions.py
bsieber-mozilla: resolved in this commit:
https://github.com/mozilla/jira-bugzilla-integration/pull/1/commits/8219d4187d7cd8cde7dd014367fdf65aa6a9d98a

src/core/config.py
bsieber-mozilla: resolved in this commit:
https://github.com/mozilla/jira-bugzilla-integration/pull/1/commits/8219d4187d7cd8cde7dd014367fdf65aa6a9d98a

pyproject.toml
```diff
@@ -0,0 +1,73 @@
...
+[tool.pytest.ini_options]
+testpaths = [
+    "tests/unit",
+]
+
+[tool.pylint]
+    [tool.pylint.'MESSAGES CONTROL']
+    disable = [
+        "C0114", #missing-module-docstring
+        "C0116", #missing-function-docstring
```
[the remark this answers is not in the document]
bsieber-mozilla: https://github.com/mozilla/jira-bugzilla-integration/issues/7

Made this an issue; I believe this can be resolved over time, and would be helpful for individuals looking to contribute to it as open source.

pyproject.toml
```diff
@@ -0,0 +1,73 @@
...
+    ]
+    [tool.pylint.REPORTS]
+    output-format = 'colorized'
+    [tool.pylint.MASTER]
+    load-plugins='pylint_pytest'
+    ignore='third_party'
+
+[tool.black]
```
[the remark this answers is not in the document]
bsieber-mozilla: https://github.com/mozilla/jira-bugzilla-integration/issues/8

pyproject.toml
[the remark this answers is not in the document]
bsieber-mozilla: https://github.com/mozilla/jira-bugzilla-integration/issues/8

.gitignore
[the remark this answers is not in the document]
bsieber-mozilla: https://github.com/mozilla/jira-bugzilla-integration/issues/8

bsieber-mozilla:
> * I'm noticing that we're using assertions in places where regular exceptions might be a better choice. I'll also point out at least one or two places where I noticed it, but I think it might be a bit of a smell to use `assert` in production code. Granted, there are places where asserts can be useful, so I'm open to pushback here, but in general it's something I (personally) would avoid using

Modified to use ValueError, ActionError, ConfigError, and ProcessErrors.

> 
> * As we kind of discussed on our call, I think some of the subpackages we're creating might not be needed just yet. Looking at it a bit more, I think all of the functions in the `core` modules could live nicely in `src.jbi.process`. I think that module has one clear job -- process the tags and actions to prepare them to be used by the FastAPI app.

I'm hesitant to re-add this together; but I will do a bit more thinking on this.

> 
> * In places where you use type hints for container objects, it might be nice to further provide the type(s) for the objects in the container. E.g. `Tuple[str, str]`

Updated where this was noticed.

----

Closing this PR as most of the review has been absorbed into #1

PR #4, opened 2022-02-22
Dockerflow and General Updates to Structure

bsieber-mozilla:


bsieber-mozilla:
Between this PR and #1; I believe this is an unnecessary one to review; it'll likely be better to "compare" #1 with the last commit from #3 to just see diffs.

PR #6, opened 2022-02-23
Log every request with MozLog

leplatrem:
r? @bsieber-mozilla

### DIFF
--- a/src/app/api.py
+++ b/src/app/api.py
@@ -1,3 +1,8 @@
+import logging
+import sys
+import time
+from datetime import datetime
+
 import uvicorn  # type: ignore
 from fastapi import FastAPI, Request
 from fastapi.responses import RedirectResponse
@@ -5,6 +10,31 @@ from fastapi.responses import RedirectResponse
 from src.app.environment import get_settings
 from src.app.monitor import api_router as monitor_router
 from src.jbi.router import api_router as jbi_router
+
+settings = get_settings()
+
+logging_config = {
+    "version": 1,
+    "formatters": {
+        "json": {
+            "()": "dockerflow.logging.JsonLogFormatter",
+            "logger_name": "jbi",
+        },
+    },
+    "handlers": {
+        "console": {
+            "level": settings.log_level.upper(),
+            "class": "logging.StreamHandler",
+            "formatter": "json",
+            "stream": sys.stdout,
+        }
+    },
+    "loggers": {
+        "request.summary": {"handlers": ["console"], "level": "INFO"},
+    },
+}
+
+logging.config.dictConfig(logging_config)
 
 app = FastAPI(
     title="Jira Bugzilla Integration (JBI)",
@@ -15,6 +45,7 @@ app = FastAPI(
 app.include_router(monitor_router)
 app.include_router(jbi_router)
 
+summary_logger = logging.getLogger("request.summary")
 
 @app.get("/", include_in_schema=False)
 def root(request: Request):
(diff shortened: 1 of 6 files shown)


src/app/api.py
```diff
@@ -6,6 +11,31 @@
...
+            "level": settings.log_level.upper(),
+            "class": "logging.StreamHandler",
+            "formatter": "json",
+            "stream": sys.stdout,
+        }
+    },
+    "loggers": {
+        "request.summary": {"handlers": ["console"], "level": "INFO"},
+    },
+}
```
leplatrem, line 35: Should this be part of the settings object instead? I've never used pydantic before

bsieber-mozilla: I don't believe this needs to be part of the settings object/pydantic; it could be but it might be better in it's own `log.py` module...

However, I do think this can be handled on the app-startup:
```
@app.on_event("startup")
def startup_event():
    configure_logging()
```

then we can call `logging.getLogger("abc.xyz")` to get the logger with the correct configuration as well as "path" _(not sure what the name of the "abc.xyz" or "request.summary" would be called here, I'm calling this "path")_

src/app/api.py
```diff
@@ -26,8 +59,32 @@ def root(request: Request):
     return RedirectResponse(url="./docs")
 
 
+@app.middleware("http")
+async def request_summary(request: Request, call_next):
+    previous_time = time.time()
+
```
bsieber-mozilla, line 65:
```suggestion
async def request_summary(request: Request, call_next):
    summary_logger = logging.getLogger("request.summary")
    
    previous_time = time.time()

```

src/app/api.py
```diff
@@ -15,6 +46,8 @@
 app.include_router(monitor_router)
 app.include_router(jbi_router)
 
+summary_logger = logging.getLogger("request.summary")
```
bsieber-mozilla, line 49:
```suggestion
```

bsieber-mozilla: suggestion: move into `request_summary` method

src/app/api.py
```diff
@@ -6,6 +12,31 @@
...
+            "formatter": "json",
+            "stream": sys.stdout,
+        }
+    },
+    "loggers": {
+        "request.summary": {"handlers": ["console"], "level": "INFO"},
+    },
+}
+
+logging.config.dictConfig(logging_config)
```
bsieber-mozilla, line 38:
```suggestion
@app.on_event("startup") 
def startup_event():
    logging_config = {
        "version": 1,
        "formatters": {
            "json": {
                "()": "dockerflow.logging.JsonLogFormatter",
                "logger_name": "jbi",
            },
        },
        "handlers": {
            "console": {
                "level": settings.log_level.upper(),
                "class": "logging.StreamHandler",
                "formatter": "json",
                "stream": sys.stdout,
            }
        },
        "loggers": {
            "request.summary": {"handlers": ["console"], "level": "INFO"},
        },
    }
    
    logging.config.dictConfig(logging_config)
```

bsieber-mozilla: Something like this maybe? Then the internals could be extracted to `log.py` and then called on this startup method.

What are your thoughts on that?

src/app/api.py
bsieber-mozilla: suggestion: move `summary_logger` into `request_summary` method

bsieber-mozilla (approved):
Thanks for this PR @leplatrem! 

I've added some comments that are more style than functionality; feel free to commit those suggestions or ignore them--we could tackle them later too.

src/app/api.py
bsieber-mozilla: _Hmm, if this "startup" event's is used; this might come in handy: https://fastapi.tiangolo.com/advanced/testing-events/_

bsieber-mozilla:
@leplatrem the suggestions presented here can be absorbed through https://github.com/leplatrem/jira-bugzilla-integration/pull/1
or that PR can be ignored and resolved at a later time.

PR #14, opened 2022-03-02
Handling some comments from previous PR; Update to use Jinja2

bsieber-mozilla:


leplatrem:
Don't you think there's a sort of redundancy between the parameters being passed to `init()` and then to the callable?

I think I would remove the latter, and let *action implementers* manage that themselves (it's relatively trivial to retain them from init(), I don't think it brings any value).

src/templates/powered_by_template.html
```diff
@@ -0,0 +1,40 @@
...
+                        <th>Enabled</th>
+                        <th>Parameters</th>
+                    </tr>
+                </thead>
+                <tbody class="list">
+                    {% for key, value in data.items() %}
+                        {% if enable_query is none or value.get("enabled") is enable_query %}
+                        <tr>
+                            <td class="identifier">{{key}}</td>
+                            <td class="action">{{value.get("action")}}</td>
```
leplatrem, line 26: nit: we know that `action` and `enabled` will be there, we can just do `value["action"]` or in Jinja I think you can also do `{{ value.action }}`

src/jbi/whiteboard_actions/README.md
```diff
@@ -1,20 +1,20 @@
...
 
 ### Create a new action...
 Let's create a `new_action`!
-1. First, add a new python file called "my_team_action" in the `src/jbi/whiteboard_actions/` dir.
+1. First, add a new python file called `my_team_action.py` in the `src/jbi/whiteboard_actions/` directory.
 1. Add the python function `init` to a new "my_team_action" module, let's use the following:
     ```python
     def init(whiteboard_tag, jira_project_key, optional_param=42):
         return lambda payload, context: print("Hi world!")
     ```
```
leplatrem, line 12: in case we keep it, the `context` is `parameters` here now

src/templates/powered_by_template.html
```diff
@@ -0,0 +1,40 @@
+<html>
+    <head>
+        <title>{{title}}</title>
+        <link rel="stylesheet" type="text/css" href="static/styles.css">
+    </head>
+    <body>
+        <h1>{{title}}</h1>
+        <p>{{num_configs}} collections.</p>
```
leplatrem, line 8: Should probably be `actions`?

leplatrem (approved):
r+wc

src/templates/powered_by_template.html
bsieber-mozilla: Yeah, that aligns better with the rest of the naming

src/jbi/whiteboard_actions/README.md
bsieber-mozilla: I like the idea of removing params

bsieber-mozilla:
> Don't you think there's a sort of redundancy between the parameters being passed to `init()` and then to the callable?
> 
> I think I would remove the latter, and let _action implementers_ manage that themselves (it's relatively trivial to retain them from init(), I don't think it brings any value).

Makes sense! Made the change.

> r+wc

not sure what you mean by this

leplatrem:
>> r+wc
>>
> not sure what you mean by this
> 

Sorry this is an old habit :grimacing: it means «approved with comments» :sweat:

leplatrem (approved):

bsieber-mozilla:
> >> r+wc
> 
> >>
> 
> > not sure what you mean by this
> 
> > 
> 
> 
> 
> Sorry this is an old habit :grimacing: it means «approved with comments» :sweat: 

Ha! I like it, esp now that I know it

PR #24, opened 2022-03-16
Fix #22: Add Sentry support

### LINKED ISSUE #22: Report errors to Sentry
We want to report stacktraces to Sentry.

Configure Sentry from a `SENTRY_DSN` env var. Skip if empty.

### LINKED ISSUE #33: Enhance the "retry" capabilities of JBI when there are external service errors
What happens here if for some reason, `jira_client.issue_add_comment` fails -- maybe there's some blip of a `500` error on Jira's end or something. Would we be in a state where only some of the comments we want on the Jira ticket appear? Should we have some sort of retry mechanism?

_Originally posted by @grahamalama in https://github.com/mozilla/jira-bugzilla-integration/pull/31#discussion_r835473860_

leplatrem:
Fix #22


tests/unit/app/test_sentry.py
```diff
@@ -0,0 +1,19 @@
...
+from unittest.mock import patch
+
+
+def test_errors_are_reported_to_sentry(anon_client):
+    with patch("src.app.api.sentry_sdk.capture_exception") as mocked:
+        with patch("src.jbi.router.execute_request", side_effect=ValueError):
+            try:
+                anon_client.post("/bugzilla_webhook")
+            except Exception:
+                pass  # This shouldn't be necessary, the API should return 500.
```
leplatrem, line 17: @bsieber-mozilla Is it expected that the server will crash on exceptions? I guess that in that case, Gunicorn (or Nginx) will be in charge of returning the 500 response. Just wanted to confirm :)

src/app/environment.py
```diff
@@ -28,6 +28,9 @@ class Settings(BaseSettings):
     # Logging
     log_level: str = "info"
 
+    # Sentry
+    sentry_dsn: str = ""
```
grahamalama, line 32: This might be a good usecase for a [URL type provided by Pydantic](https://pydantic-docs.helpmanual.io/usage/types/#urls). Also, I think it could be `Optional[<url type>]` since that's what the [Sentry client accepts](https://github.com/getsentry/sentry-python/blob/de0bc5019c715ecbb2409a852037530f36255d75/sentry_sdk/consts.py#L49).

tests/unit/app/test_sentry.py
bsieber-mozilla: On exceptions there should be a 500 response without crashing the server; I checked out the branch to test this and confirmed.

leplatrem: I don't say that the server crashes. But the application does. 
Gunicorn catches the crash and returns a 500. 

It's totally acceptable, we just have to take it into account in unit tests, since we don't test the app through gunicorn

src/app/api.py
```diff
@@ -61,10 +62,34 @@ async def request_summary(request: Request, call_next):
...
+        raise exc
+
+
 @app.on_event("startup")
 def startup_event():
     """On app startup perform these setup operations"""
     configure_logging()
+    sentry_sdk.init(  # pylint: disable=abstract-class-instantiated  # noqa: E0110
+        dsn=settings.sentry_dsn
+    )
```
bsieber-mozilla, line 92: [This comment](https://github.com/mozilla-it/ctms-api/blob/527d21d45922b3361325319487ab07254739a4eb/ctms/exception_capture.py#L11-L20) from the CTMS/FastAPI interactions with Sentry is a bit confusing to me. We may need this sentry.init to happen elsewhere--if I'm reading that comment correctly.

We can try this as is and make modifications if needed

leplatrem: This startup callback was something you suggested [here](https://github.com/mozilla/jira-bugzilla-integration/pull/6#discussion_r813419676), but maybe it's simpler to do it during the app instantiation.

bsieber-mozilla: Yeah, I think this is still the right location to startup the logging; but in CTMS--that comment linked above makes me wonder how threadsafe it is to do the actual `init` in this callback for sentry. 

In CTMS, we just add it to the app like this:
https://github.com/mozilla-it/ctms-api/blob/527d21d45922b3361325319487ab07254739a4eb/ctms/app.py#L107-L110

```    
init_sentry()
app.add_middleware(SentryAsgiMiddleware) ```

bsieber-mozilla: In CTMS we only migrated Sentry out of the startup callback:

https://github.com/mozilla-it/ctms-api/blob/527d21d45922b3361325319487ab07254739a4eb/ctms/app.py#L113-L120

bsieber-mozilla: I'm ok if we choose to resolve this comment; we can address whether we want to use the 'startup' callback in a later PR. 

(I'm still of the mind that Sentry is the outlier with this issue of using the startup callback.)

bsieber-mozilla:
Do we have the Sentry DSN setup?

PR #25, opened 2022-03-16
Add proof-of-concept ProjectFlow lint action

leplatrem:


.github/workflows/projectflow-lint.yaml
```diff
@@ -0,0 +1,9 @@
+on: [push]
```
bsieber-mozilla, line 1:
```suggestion
on: [push, pull_request]
```

bsieber-mozilla (commented):
What is the intent of this as a linter? (Common project "expectations"?)

It would likely not overlap with our current linters; we've got a few linters currently:
- [pre-commit](https://github.com/mozilla/jira-bugzilla-integration/blob/main/.pre-commit-config.yaml)
- [docker `lint` target](https://github.com/mozilla/jira-bugzilla-integration/blob/d5f84bc90b5e4fe1fd3172f71c9df38b5b3a901b/infra/Dockerfile#L83-L92)
- [yamllint](https://github.com/mozilla/jira-bugzilla-integration/blob/main/.yamllint)

leplatrem:
> What is the intent of this as a linter? (Common project "expectations"?)

This would have been a way to easily lint any repo, just by adding the action :) But I was just trying things out... will come back to this!

Issue #20, opened 2022-03-12
Migrate Default's into Action Configuration YAML?


bsieber-mozilla:
The following lines establish defaults for a a few of the action config fields:
https://github.com/mozilla/jira-bugzilla-integration/blob/d5f84bc90b5e4fe1fd3172f71c9df38b5b3a901b/src/jbi/configuration.py#L47-L49

Perhaps a new section in the config.*.yaml could be added to control this from configuration.
https://github.com/mozilla/jira-bugzilla-integration/blob/d5f84bc90b5e4fe1fd3172f71c9df38b5b3a901b/config/config.dev.yaml#L1-L5

_Should there be a new section? Maybe there's an alternate route to explore here_

Should this new section be: `default`, `default.actions`, `actions._default`, or `actions.default`.
- [ ] `default`: as a top-level section
- [ ] `default.actions`: a top-level section that has a section dedicated to the other top-level section `action`
- [ ] `actions._default`: adding the underscore should ensure yamllint keeps this element at the top of the list (we can also remove ordering of keys from yamllint config)
- [ ] `actions.default` 🤷

bsieber-mozilla:
Not something we need to act on at this time; but--time-permitting--I'd love your thoughts on this: @grahamalama @leplatrem

grahamalama:
This might also be a thing we could use pydantic for:
```python
class ActionConfig(BaseModel, extra=Extra.allow):
    action: str = "src.jbi.whiteboard_actions.default"
    enabled: bool = False
    parameters: dict = {}
```

leplatrem:
I like the idea to rely on pydantic models

leplatrem:
Implemented in #28

PR #28, opened 2022-03-18
Using pydantic-yaml and validations through pydantic.validators.

bsieber-mozilla:
This PR address issue #20


src/templates/powered_by_template.html
```diff
@@ -20,7 +20,7 @@ <h1>{{title}}</h1>
                 </thead>
                 <tbody class="list">
                     {% for key, value in data.items() %}
-                        {% if enable_query is none or value.enabled is enable_query %}
+                        {% if enable_query == none or value.enabled == enable_query %}
```
leplatrem, line 23: Why this change?

src/jbi/configuration.py
```diff
@@ -1,83 +1,33 @@
...
-    """Errors used to determine when the Configuration is invalid"""
-
-
-class ProcessError(Exception):
-    """Error when an exception occurs during processing"""
+    """Error when an exception occurs during processing config"""
 
 
+@lru_cache
 def get_yaml_configurations(
```
leplatrem, line 21: I think it would be clearer if we would use the term *actions configuration* throughout the whole code base, instead of *yaml configuration*.

For example, this function could be called `get_actions()`, especially since it's in the configuration module. 

`actions = configuration.get_actions()`

src/jbi/model.py
```diff
@@ -0,0 +1,64 @@
...
+from types import ModuleType
+from typing import Any, Dict, Optional
+
+from pydantic import Extra, ValidationError, root_validator, validator
+from pydantic_yaml import YamlModel
+
+
+class ActionConfig(YamlModel, extra=Extra.allow):
+    """
+    ActionConfig is the inner model of `actions` in yaml
```
leplatrem, line 15: This comment is a bit too similar to the one of `Actions`.

```suggestion
    ActionConfig is the inner model for each action in the configuration file
```

src/jbi/model.py
```diff
@@ -0,0 +1,64 @@
...
+        except ImportError as exception:
+            raise ValidationError(f"unknown action `{action}`.") from exception
+        except (TypeError, AttributeError) as exception:
+            raise ValidationError("action is not properly setup.") from exception
+        return values
+
+
+class Actions(YamlModel):
+    """
+    Actions is the overall model from parsing the yaml config `actions` sections
```
leplatrem, line 46:
```suggestion
    Actions is the overall model for the list of `actions` in the configuration file
```

src/jbi/model.py
```diff
@@ -0,0 +1,64 @@
...
+import importlib
+from inspect import signature
+from types import ModuleType
+from typing import Any, Dict, Optional
+
+from pydantic import Extra, ValidationError, root_validator, validator
+from pydantic_yaml import YamlModel
+
+
+class ActionConfig(YamlModel, extra=Extra.allow):
```
leplatrem, line 13: Shouldn't it be just `Action` (since we have `Actions` below) ?

src/jbi/model.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Actions(YamlModel):
+    """
+    Actions is the overall model from parsing the yaml config `actions` sections
+    """
+
+    actions: Dict[str, ActionConfig]
+
+    @validator("actions")
+    def validate_action_yaml_jbi_naming(
```
leplatrem, line 52:
```suggestion
    def validate_action_matches_whiteboard(
```

src/jbi/router.py
```diff
@@ -36,7 +36,7 @@ def get_whiteboard_tag(
     whiteboard_tag: Optional[str] = None,
 ):
     """API for viewing whiteboard_tags"""
-    data = configuration.get_yaml_configurations()
+    data = configuration.get_yaml_configurations().dict()["actions"]
```
leplatrem, line 39: Maybe this helper on the configuration module should return the dict directly

src/templates/powered_by_template.html
bsieber-mozilla: https://github.com/pallets/jinja/issues/520#issuecomment-164306945
`is` is a keyword--I was running into 500s when using the dev config and modifiying the query param; so went with `==`

src/jbi/model.py
bsieber-mozilla: That works for me. I think the naming has evolved over the time--and I might rename the config files to action.*.yaml as well

src/jbi/router.py
bsieber-mozilla: I did that initially but I was wondering whether we want to be using objects or dictionary's to access the underlying data. 

I can migrate this to a helper for the time being.

bsieber-mozilla:
Thanks for your review @leplatrem I made the suggested changes in the most recent commit

src/jbi/configuration.py
```diff
@@ -1,83 +1,38 @@
...
-
-
-def get_yaml_configurations(
+@lru_cache
+def get_actions(
     jbi_config_file: str = f"config/config.{settings.env}.yaml",
-) -> Dict[str, Dict]:
-    """Convert and validate YAML configuration to python dict"""
+) -> Actions:
+    """Convert and validate YAML configuration to Action object"""
```
leplatrem, line 24:
```suggestion
    """Convert and validate YAML configuration to `Action` objects"""
```

src/jbi/configuration.py
```diff
@@ -1,83 +1,38 @@
...
-        if not hasattr(action_module, "init"):
-            raise TypeError("Module is missing `init` method.")
 
-        signature(action_module.init).bind(**action_parameters)  # type: ignore
-    except ImportError as exception:
-        raise ConfigError(f"Unknown action `{action}`.") from exception
-    except (TypeError, AttributeError) as exception:
-        raise ConfigError("Action is not properly setup.") from exception
+def get_actions_dict():
+    """Returns dict of"""
```
leplatrem, line 37:
```suggestion
    """Returns dict of `get_actions()`"""
```

leplatrem (approved):

leplatrem (approved):

PR #31, opened 2022-03-24
Adding Default Action

bsieber-mozilla:


src/jbi/whiteboard_actions/default.py
```diff
@@ -31,10 +37,111 @@ def __init__(self, **kwargs):
...
+                    jira_update_response = jira_client.update_issue_field(
+                        key=linked_issue_key, fields=fields
+                    )
+                    # comment
+                    jira_comment_responses = []
+                    for comment in comments:
+                        jira_comment_responses.append(
+                            jira_client.issue_add_comment(
+                                issue_key=linked_issue_key, comment=comment
+                            )
```
grahamalama, line 86: What happens here if for some reason, `jira_client.issue_add_comment` fails -- maybe there's some blip of a `500` error on Jira's end or something. Would we be in a state where only some of the comments we want on the Jira ticket appear? Should we have some sort of retry mechanism?

src/jbi/whiteboard_actions/default.py
```diff
@@ -31,10 +37,111 @@ def __init__(self, **kwargs):
...
+                if isinstance(response, list):
+                    response = response[0]
+
+                if isinstance(response, dict):
+                    if any(
+                        element in ["errors", "errorMessages"] and response[element]
+                        for element in response.keys()
+                    ):
+                        # Failure to create: err keys exist with value
+                        return None
```
grahamalama, line 111: Will this return some 500 error from our server? Should we handle this and surface whatever error messages Jira send back?

tests/unit/jbi/whiteboard_actions/test_default.py
```diff
@@ -19,10 +22,34 @@ def test_default_returns_callable_without_data():
...
+        return {}
+
+    monkeypatch.setattr(bugzilla.Bugzilla, "getbug", mock_getbug)
+    monkeypatch.setattr(bugzilla.Bugzilla, "update_bugs", mock_update_bugs)
+    monkeypatch.setattr(Jira, "create_issue", mock_create_issue)
+    monkeypatch.setattr(
+        Jira,
+        "create_or_update_issue_remote_links",
+        mock_create_or_update_issue_remote_links,
+    )
```
grahamalama, line 51: If we have to do many more tests with these services, monkeypatching for each could become cumbersome. Not something to handle in this PR, but it _might_ be worth exploring a dependency injection approach for requiring different services in actions. That way, we could provide some `MockBugzillaClient` or `MockJiraClient` to the actions when we test them.

src/jbi/router.py
```diff
@@ -1,34 +1,87 @@
...
+        action_module: ModuleType = importlib.import_module(current_action["action"])
+        if not action_module:
+            raise ValidationError("action not found")
+
+        callable_action = action_module.init(  # type: ignore
+            **current_action["parameters"]
+        )
+        return callable_action(payload=request)
+    except ValidationError as exception:
+        return JSONResponse(content={"error": str(exception)}, status_code=201)
```
grahamalama, line 71: Should we use some `4xx` error here?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def is_private_comment(self) -> bool:
+        """Helper function to determine if this comment private--not accessible or open"""
+        if self.is_private is None:
+            self.is_private = False
+        return self.is_private
+
+
+class BugzillaBug(BaseModel):
+    """Bugzilla Bug Object"""
```
grahamalama, line 81: Is there Bugzilla documentation we can link to here? Might be a useful reference in the future.

grahamalama (commented):
A few inline comments as a first pass, but looking good so far.

src/jbi/whiteboard_actions/default.py
```diff
@@ -31,10 +37,111 @@ def __init__(self, **kwargs):
...
+                        "jira_response": jira_response,
+                    }
+                    return JSONResponse(content=content, status_code=201)
+                # else:
+                response = jira_client.delete_issue(
+                    issue_id_or_key=jira_key_in_response
+                )
+                return JSONResponse(content={"response": response}, status_code=201)
+        except ValidationError as exception:
+            return JSONResponse(content={"error": exception}, status_code=201)
```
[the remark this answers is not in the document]
bsieber-mozilla: Any other exceptions would result in a 5XX and would be retried; these validations exceptions are "confirmed" to the sender (bugzilla webhook push server) as a 200/201 so that the event queue can release them and continue.

src/jbi/whiteboard_actions/default.py
```diff
@@ -31,10 +37,111 @@ def __init__(self, **kwargs):
         self.whiteboard_tag = kwargs.get("whiteboard_tag")
         self.jira_project_key = kwargs.get("jira_project_key")
 
-    def __call__(self, payload):
+    def __call__(  # pylint: disable=too-many-locals,too-many-return-statements,inconsistent-return-statements
```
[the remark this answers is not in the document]
bsieber-mozilla: Yeah, that makes sense; splitting the helpers out will also help with documentation--like you mentioned in another comment.

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class BugzillaWebhookEvent(BaseModel):
+    """Bugzilla Event Object"""
+
+    action: str
+    time: Optional[datetime.datetime] = None
+    user: Optional[BugzillaWebhookUser] = None
+    changes: Optional[List[BugzillaWebhookEventChange]] = None
+    target: Optional[str] = None
```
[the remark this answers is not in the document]
bsieber-mozilla: Good question; I believe the `= None` was because of FastAPI's use of Optional's and how to present them in the swagger docs. Since we're not using it as a Response, Request, Body, Query, or Path object in a FastAPI sense; then I believe the `= None` is not a requirement of Pydantic and is just additional bloat-text.

src/jbi/whiteboard_actions/default.py
bsieber-mozilla: Currently, the behavior is: let the message go, have user re-trigger through bump/minor-update on card. If issues still occur seek manual intervention.

There's some additional handling for retries _we might_ want to consider adding here post-launch; I think we can continue the current process. 

_(Historical: Initially, in v1, there were an attempts at retries; this resulted in some inconsistencies with duplicate creations and duplicate jira issues--with and without tags. There's gonna be some finesse required for this.)_

src/jbi/whiteboard_actions/default.py
bsieber-mozilla: > Would we be in a state where only some of the comments we want on the Jira ticket appear? 

Currently that is the case.

> Should we have some sort of retry mechanism?

I think we may want to reconsider "retries" or perhaps complete bug validations (confirm bug is 1:1 with issue--comments, values etc) whenever we receive an event.

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+        if not payload.bug:
+            raise ActionError("payload does not contain bug data")
+        current_bug_info = self.bugzilla_client.getbug(payload.bug.id)
+        bug_obj = BugzillaBug.parse_obj(current_bug_info.__dict__)
+        jira_key_in_bugzilla = bug_obj.extract_from_see_also()
+        _duplicate_creation_event = (
+            jira_key_in_bugzilla is not None
+            and jira_key_in_response != jira_key_in_bugzilla
+        )
+        if not _duplicate_creation_event:
```
leplatrem, line 125: Since the block in the case `_duplicate_creation_event` is shorter than the block `not _duplicate_creation_event`, I would invert it here. This way the long block below won't be indented and understood as the main code path

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
-    def __call__(self, payload):
-        """
-        Called from BZ webhook
-        Call Jira SDK with project key etc.
-        """
-        print(payload)
-        print(self.parameters)
+            bugzilla_url = (
+                f"{self.settings.bugzilla_base_url}show_bug.cgi?id={bug_obj.id}"
+            )
```
leplatrem, line 132: `BUGZILLA_URL = f"{self.settings.bugzilla_base_url}show_bug.cgi?id=%s"`

...

`bugzilla_url = BUGZILLA_URL.format(bug_obj.id)`

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,247 @@
...
+    id: Optional[int]
+    number: Optional[int]
+    is_private: Optional[bool]
+    creation_time: Optional[datetime.datetime]
+
+    def is_comment_description(self) -> bool:
+        """Used to determine if `self` is a description or comment."""
+        if self.number in [0]:  # Currently 0th comment is the description
+            return True
+        return False
```
leplatrem, line 68:
```suggestion
        return self.number == 0  # Currently 0th comment is the description
```

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,247 @@
...
+    def is_comment_generic(self) -> bool:
+        """All comments after comment-0 are generic"""
+        is_description = self.is_comment_description()
+        return not is_description
+
+    def is_private_comment(self) -> bool:
+        """Helper function to determine if this comment private--not accessible or open"""
+        if self.is_private is None:
+            self.is_private = False
+        return self.is_private
```
leplatrem, line 79:
```suggestion
        return bool(self.is_private)
```

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def get_whiteboard_with_brackets_as_list(self):
+        """Convert string whiteboard into list, splitting on ']' and removing '['; then re-adding."""
+        wb_list = self.get_whiteboard_as_list()
+        if wb_list is not None and len(wb_list) > 0:
+            return [f"[{element}]" for element in wb_list]
+        return []
+
+    def get_jbi_labels(self):
+        """bugzilla is a required label for SYNC project and is being used to prevent gh syncing"""
```
leplatrem, line 124: gh syncing?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,247 @@
...
+            try:
+                parsed_url: ParseResult = urlparse(url=url)
+                expected_hosts = ["jira", "atlassian"]
+
+                if any(  # pylint: disable=use-a-generator
+                    [part in expected_hosts for part in parsed_url.hostname.split(".")]
+                ):
+                    parsed_jira_key = parsed_url.path.split("/")[-1]
+                    return parsed_jira_key
+            except Exception:  # pylint: disable=broad-except
```
leplatrem, line 175: Can't we list exceptions here?

src/jbi/router.py
```diff
@@ -1,34 +1,84 @@
...
+        bugzilla_client = get_bugzilla()
+        current_bug_info = bugzilla_client.getbug(request.bug.id)
+        bug_obj = BugzillaBug.parse_obj(current_bug_info.__dict__)
+        current_action = extract_current_action(bug_obj, action_map, settings)
+        if not current_action:
+            raise IgnoreInvalidRequestError("bug does not have matching config")
+
+        action_module: ModuleType = importlib.import_module(current_action["action"])
+        if not action_module:
+            raise IgnoreInvalidRequestError("action not found")
```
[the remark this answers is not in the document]
bsieber-mozilla: I believe this was added as pre-commit/mypy wasn't happy about the potential of action_module being "Nonetype" when calling `init` in the following line. This `if` was added as an additional confirmation to avoid the "Nonetype" issue when calling `init`; however I can add an ignore for mypy instead and let it error otherwise--which would be unexpected at this point in the code anyway.

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,247 @@
...
+    def get_potential_whiteboard_config_list(self):
+        """Get all possible whiteboard_tag configuration values"""
+        converted_list: List = []
+        for whiteboard in self.get_whiteboard_as_list():
+            converted_tag = self.convert_wb_to_tag(whiteboard=whiteboard)
+            if converted_tag not in [None, "", " "]:
+                converted_list.append(converted_tag)
+
+        if len(converted_list) == 0:
+            return None
```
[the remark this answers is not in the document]
bsieber-mozilla: Thanks for the suggestion. Updated in most recent commit!

src/jbi/router.py
```diff
@@ -1,34 +1,84 @@
...
-def execute_request(request, settings):
+def extract_current_action(  # pylint: disable=inconsistent-return-statements
+    bug_obj: BugzillaBug, action_map, settings
+):
+    """Find first matching action from bug's whiteboard list"""
+    potential_configuration_tags: List[
+        str
+    ] = bug_obj.get_potential_whiteboard_config_list()
+    if not potential_configuration_tags:
+        return {}
```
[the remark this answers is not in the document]
bsieber-mozilla: Thanks for the suggestion. Updated in most recent commit: no longer returning empty dict!

src/jbi/router.py
```diff
@@ -1,34 +1,84 @@
...
+        is_private_bug = request.bug.is_private
+        if is_private_bug:
+            raise IgnoreInvalidRequestError("private bugs are not valid")
+
+        bugzilla_client = get_bugzilla()
+        current_bug_info = bugzilla_client.getbug(request.bug.id)
+        bug_obj = BugzillaBug.parse_obj(current_bug_info.__dict__)
+        current_action = extract_current_action(bug_obj, action_map, settings)
+        if not current_action:
+            raise IgnoreInvalidRequestError("bug does not have matching config")
```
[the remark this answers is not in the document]
bsieber-mozilla: Thanks for the suggestion. Updated in most recent commit!

src/jbi/router.py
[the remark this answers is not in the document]
bsieber-mozilla: I've updated this in the most recent commit.

src/jbi/router.py
```diff
@@ -1,34 +1,84 @@
...
+            raise IgnoreInvalidRequestError("action not found")
+
+        callable_action = action_module.init(  # type: ignore
+            **current_action["parameters"]
+        )
+        return callable_action(payload=request)
+    except (IgnoreInvalidRequestError, ActionError) as exception:
+        return JSONResponse(content={"error": str(exception)}, status_code=201)
+    except Exception:  # pylint: disable=broad-except
+        return JSONResponse(content={"error": traceback.format_exc()}, status_code=500)
```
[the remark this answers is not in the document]
bsieber-mozilla: Done!

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+        self.jira_client = get_jira()
+        self.settings = get_settings()
+
+    def __call__(  # pylint: disable=too-many-locals,too-many-return-statements,inconsistent-return-statements
+        self, payload
+    ):
+        """Called from BZ webhook when default action is used. All default-action webhook-events are processed here."""
+        payload: BugzillaWebhookRequest = BugzillaWebhookRequest.parse_obj(
+            payload
+        )  # typing assistance
```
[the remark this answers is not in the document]
bsieber-mozilla: I wasn't sure if we'd want to do it for all actions; but I've gone ahead and done that. The user's of custom actions can always convert it to a dict if they prefer.

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+
+        current_bug_info = self.bugzilla_client.getbug(payload.bug.id)
+        bug_obj = BugzillaBug.parse_obj(current_bug_info.__dict__)
```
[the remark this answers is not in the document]
bsieber-mozilla: Made this change, thanks for the suggestion!

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+        target = payload.event.target
+        linked_issue_key = bug_obj.extract_from_see_also()
+
+        if target == "comment":
+            self.event_target_comment(
```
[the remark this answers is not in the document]
bsieber-mozilla: I've updated the names; hopefully they make a bit more sense now!

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+        if target == "bug":
+            self.event_target_bug(
+                payload=payload, linked_issue_key=linked_issue_key, bug_obj=bug_obj
+            )
+
+    def event_target_comment(self, payload: BugzillaWebhookRequest, linked_issue_key):
+        """Confirm issue is already linked, then apply comments; otherwise noop"""
+        if not linked_issue_key:
+            # noop
+            return JSONResponse(content={"status": "noop"}, status_code=201)
```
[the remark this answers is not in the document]
bsieber-mozilla: Agreed; removed responses from action logic--a dict returned indicates the content for a response otherwise an error should be thrown.

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+                status_code=201,
+            )
+        # else: create jira issue
+        fields = bug_obj.get_jira_issue_dict()
+        fields["project"] = {"key": self.jira_project_key}
+
+        response = self.jira_client.create_issue(fields=fields)
+
+        if isinstance(response, list):
+            response = response[0]
```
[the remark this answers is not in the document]
bsieber-mozilla: Added some comments

src/jbi/whiteboard_actions/default.py
bsieber-mozilla: Updated!

src/jbi/whiteboard_actions/default.py
bsieber-mozilla: Moved to settings object

src/jbi/bugzilla_objects.py
bsieber-mozilla: Removed this message; it's a requirement by the Sync team--they have two syncing solutions github issues and jbi, they use this "bugzilla" tag to determine the origin of the issue for queries in jira

src/jbi/bugzilla_objects.py
bsieber-mozilla: We could, I believe there's a possibility of an AttributeError; I'm not sure all the possibilities, but I would feel more comfortable leaving this as is _for now._

src/jbi/router.py
bsieber-mozilla: The reason we don't use 4xx here is because these Errors are more of a Filtration component than an "error"
when these "invalid" requests are received we want to just move on and get the next message in the bugzilla webhook queue--a 4xx results in bugzilla retrying the request, since we want it "dropped" from the queue we need to give it a "success" since it was received and "processed".

_Perhaps using error for this is not appropriate..._

src/jbi/router.py
```diff
@@ -1,34 +1,79 @@
...
+    bug_obj: BugzillaBug, action_map, settings
+):
+    """Find first matching action from bug's whiteboard list"""
+    potential_configuration_tags: List[
+        str
+    ] = bug_obj.get_potential_whiteboard_config_list()
+
+    for tag in potential_configuration_tags:
+        if tag.lower() in action_map:
+            return action_map.get(tag.lower())
```
leplatrem, line 36: Actually, this would be a use-case for `:=` :) 

```suggestion
        if action := action_map.get(tag.lower()):
            return action
```

src/jbi/router.py
leplatrem: I'm also doubtful about that. 

Are you sure Bugzilla retries on 4XX? 

Retrying by sending the same request twice would make sense for a 5XX response. But for a 4XX response, the client has to retry after having changed the request. 

In any case, 201 is not the appropriate code, since it means «successfully created».

Also, if an action raises an `ActionError` (because JIRA is down or whatever), then we want it to be retried by Bugzilla.

src/jbi/whiteboard_actions/default.py
```diff
@@ -6,35 +6,130 @@
...
+        self.jira_client = get_jira()
+        self.settings = get_settings()
+
+    def __call__(  # pylint: disable=too-many-locals,too-many-return-statements,inconsistent-return-statements
+        self, payload: BugzillaWebhookRequest
+    ):
+        """Called from BZ webhook when default action is used. All default-action webhook-events are processed here."""
+        target = payload.event.target
+        if not payload.bug or not isinstance(payload.bug, BugzillaBug):
+            raise ActionError("payload is expected to have bug data")
```
leplatrem, line 40: I believe that you can consider here that `payload.bug` is set. It could be part of the «contract» that the caller has with an action executer. In other words, if we garantee it at the caller level, it does not have to be checked in every action implementation

src/jbi/whiteboard_actions/default.py
```diff
@@ -6,35 +6,130 @@
...
+            # if a dict is returned or the first item in a list, confirm there are no errors
+            if any(
+                element in ["errors", "errorMessages"] and response[element]
+                for element in response.keys()
+            ):
+                raise ActionError(f"response contains error: {response}")
+
+        jira_key_in_response = response.get("key")
+        if not payload.bug:
+            raise ActionError("payload does not contain bug data")
```
leplatrem, line 108: You shouldn't have to test this again here.

src/jbi/whiteboard_actions/default.py
```diff
@@ -5,36 +5,141 @@
...
+                for element in response.keys()
+            ):
+                # Failure to create: err keys exist with value
+                raise ActionError(f"response contains error: {response}")
+
+        jira_key_in_response = response.get("key")
+        if not payload.bug:
+            raise ActionError("payload does not contain bug data")
+        current_bug_info = self.bugzilla_client.getbug(payload.bug.id)
+        bug_obj = BugzillaBug.parse_obj(current_bug_info.__dict__)
```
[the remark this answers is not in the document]
leplatrem: You already have the `bug_obj`!

src/jbi/whiteboard_actions/default.py
```diff
@@ -6,35 +6,130 @@
...
+    ):
+        """Called from BZ webhook when default action is used. All default-action webhook-events are processed here."""
+        target = payload.event.target
+        if not payload.bug or not isinstance(payload.bug, BugzillaBug):
+            raise ActionError("payload is expected to have bug data")
+
+        if target == "comment":
+            self.comment_create_or_noop(payload=payload)
+        if target == "bug":
+            self.bug_create_or_update(payload=payload)
```
leplatrem, line 45: Something seems wrong here since you don't use the returned values of these functions

src/jbi/whiteboard_actions/default.py
```diff
@@ -6,35 +6,130 @@
...
+        jira_key_in_bugzilla = bug_obj.extract_from_see_also()
+        _duplicate_creation_event = (
+            jira_key_in_bugzilla is not None
+            and jira_key_in_response != jira_key_in_bugzilla
+        )
+        if _duplicate_creation_event:
+            response = self.jira_client.delete_issue(
+                issue_id_or_key=jira_key_in_response
+            )
+            return {"response": response}
```
leplatrem, line 120: Why not `jira_response` here? 

Also, in `comment_create_or_noop()`  you return a  `status` key, and you don't here. It would be good to have some convention between the two, so that `__call__()` has consistent return values

bsieber-mozilla:
Thanks @leplatrem; I've resolved the comments in the most recent commit I believe.

src/jbi/bugzilla_objects.py
leplatrem: We don't even want to log anything?

src/jbi/router.py
```diff
@@ -1,34 +1,79 @@
...
 
-from fastapi import APIRouter, Depends, Request
-from fastapi.responses import HTMLResponse
+from fastapi import APIRouter, Body, Depends, Request
+from fastapi.responses import HTMLResponse, JSONResponse
 from fastapi.templating import Jinja2Templates
 
 from src.app import environment
 from src.jbi import configuration
+from src.jbi.bugzilla_objects import BugzillaBug, BugzillaWebhookRequest
```
leplatrem, line 15: Maybe it could just be `from src.jbi.bugzilla import BugzillaBug, BugzillaWebhookRequest` ?

The `_objects` does not feel very pythonic

tests/unit/jbi/whiteboard_actions/test_default.py
```diff
@@ -19,10 +22,34 @@ def test_default_returns_callable_without_data():
...
     try:
-        callable_object(payload={})
+        callable_object(payload=webhook_request_example)
     except Exception as exception:  # pylint: disable=broad-except
         assert False, f"`default` raised an exception {exception}"
```
leplatrem, line 55: Why don't we just have `callable_object(payload=webhook_request_example)`? Without the `except` and `assert False` ?

tests/unit/jbi/whiteboard_actions/test_default.py
```diff
@@ -19,10 +22,34 @@ def test_default_returns_callable_without_data():
...
+    def mock_create_issue(self, fields):
+        return {}
+
+    def mock_create_or_update_issue_remote_links(self, issue_key, link_url, title):
+        return {}
+
+    monkeypatch.setattr(bugzilla.Bugzilla, "getbug", mock_getbug)
+    monkeypatch.setattr(bugzilla.Bugzilla, "update_bugs", mock_update_bugs)
+    monkeypatch.setattr(Jira, "create_issue", mock_create_issue)
+    monkeypatch.setattr(
```
leplatrem, line 47: I think that this is a bit weird. 

I have the feeling that you can do much simler with `unittest.mock` (*not tested*):

```py
from unittest import mock

def test_default_returns_callable_with_data(webhook_request_example):
    with mock.patch("src.jbi.whiteboard_actions.default.get_jira") as mocked_jira:
        with mock.patch("src.jbi.whiteboard_actions.default.get_bugzilla") as mocked_bz:
            mocked_bz.getbug.return_value = webhook_request_example.bug
            value = callable_object(payload=webhook_request_example)

    assert value["status"] == "..."   
```

src/jbi/router.py
```diff
@@ -1,34 +1,79 @@
...
+from src.jbi.model import ActionError, IgnoreInvalidRequestError
```
leplatrem, line 16: Shouldn't `model` be plural? `models`? And shouldn't errors go to their own `errors` module?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+    resolution: Optional[str]
+    see_also: Optional[List]
+    summary: Optional[str]
+    severity: Optional[str]
+    priority: Optional[str]
+    creator: Optional[str]
+    assigned_to: Optional[str]
+    comment: Optional[BugzillaWebhookComment]
+
+    def get_whiteboard_count(self):
```
leplatrem, line 99: This method is not used and can probably be removed.

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+        """Convert string whiteboard into list, splitting on ']' and removing '['; then re-adding."""
+        wb_list = self.get_whiteboard_as_list()
+        if wb_list is not None and len(wb_list) > 0:
+            return [f"[{element}]" for element in wb_list]
+        return []
+
+    def get_jbi_labels(self):
+        """
+        whiteboard labels are added as a convenience for users to search in jira;
+        bugzilla is a required label for SYNC project
```
leplatrem, line 122: What does `for SYNC project` mean?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+        return []
+
+    def get_whiteboard_with_brackets_as_list(self):
+        """Convert string whiteboard into list, splitting on ']' and removing '['; then re-adding."""
+        wb_list = self.get_whiteboard_as_list()
+        if wb_list is not None and len(wb_list) > 0:
+            return [f"[{element}]" for element in wb_list]
+        return []
+
+    def get_jbi_labels(self):
```
leplatrem, line 119: I would call this `get_jira_labels()`

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+        converted_list: List = []
+        for whiteboard in self.get_whiteboard_as_list():
+            converted_tag = self.convert_wb_to_tag(whiteboard=whiteboard)
+            if converted_tag not in [None, "", " "]:
+                converted_list.append(converted_tag)
+
+        return converted_list
+
+    @staticmethod
+    def convert_wb_to_tag(whiteboard):
```
leplatrem, line 141: Why is this static? Do we call this from elsewhere?

Also, other methods have `whiteboard` in their name, why not this one?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+        """Extract tag from whiteboard label"""
+        _exists = whiteboard not in (" ", "")
+        if not _exists:
+            return ""
+        return whiteboard.split(sep="-", maxsplit=1)[0].lower()
+
+    def get_jira_issue_dict(self):
```
leplatrem, line 148: To match other methods, could this be called `map_as_jira_issue()`?

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+    def map_as_tuple_of_field_dict_and_comments(
+        self,
+        status_log_enabled: bool = True,
+        assignee_log_enabled: bool = True,
+    ) -> Tuple[Dict, List]:
+        """Extract update dict and comment list from Webhook Event"""
+
+        comments: List = []
+        if not self.bug:
+            return {}, []
```
leplatrem, line 210: Don't we want to raise a ValueError explicit here? It feels dangerous to have this silent implicit conversion to empty objects

src/jbi/bugzilla_objects.py
```diff
@@ -0,0 +1,244 @@
...
+        if self.event.changes:
+            for change in self.event.changes:
+
+                if status_log_enabled and change.field in ["status", "resolution"]:
+                    comment_dict = {
+                        "modified by": "unknown",
+                        "resolution": bug.resolution,
+                        "status": bug.status,
+                    }
+                    if self.event.user:
```
leplatrem, line 226: I would move this out of the loop, on top:
```py
user = self.event.user.login if self.event.user else "unknown"
```

leplatrem (commented):
There are a couple of things that can still be improved I think, if you still have some forces ;) 

- the fact that we return 201 on errors. See my comment https://github.com/mozilla/jira-bugzilla-integration/pull/31/files#r838330014
- the fact that we refetch/parse the bug in the action code https://github.com/mozilla/jira-bugzilla-integration/pull/31/files#r838340107
- the way jira and bugzilla are mocked in tests

We've never been so close ;)

src/jbi/router.py
bsieber-mozilla:
```Errors

If the response is not 200, the Bugzilla system will retry 4 attempts, waiting 5s, 10s, 15s and 20s between each attempt. If none of those attempts is successful, the system will continue trying every 15 minutes.

If a message is stuck without a successful attempt, the next messages that trigger the webhook will be stored in a queue in the order that were triggered and will be delivered in that order when the first message in the queue is delivered.```

https://bugzilla.mozilla.org/page.cgi?id=webhooks.html

Comment (bsieber-mozilla) [reply src/jbi/router.py]: Perhaps [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) then?

File: src/jbi/whiteboard_actions/default.py (hunk 28 above)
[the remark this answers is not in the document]
Comment (bsieber-mozilla) [reply src/jbi/whiteboard_actions/default.py:106]: We do, however that was at the time of receiving the request...This is an additional call to bugzilla to get the bug as of this time of processing(just after a jira issue create)... given that multiple messages could be handled at once, we want to re-call bugzilla to ensure the jira issue we just added added is the only one.

(We had issues with simultaneous duplicates at nearly the same second--off by microseconds. Which the system had trouble capturing.)

I'm hesitant to remove it, but maybe this is no longer necessary.

File: src/jbi/bugzilla_objects.py (hunk 36 above)
Comment (bsieber-mozilla) [reply src/jbi/bugzilla_objects.py]: modified comment

**Review by bsieber-mozilla — COMMENTED** [review]
Thanks again @leplatrem for your review; I've added some comments--and willing to further adjust, but I think most of the comments have been resolved in the most recent commit.

**Review by leplatrem — COMMENTED** [review]
:100: 

The 202 status is fine yes. What was wondering me most was the *catch all* of action errors.

The only thing left is the logging. I'm a little bit scared to have requests ignored silently, especially for the first cycles of deployment.

**Review by bsieber-mozilla — COMMENTED** [review]
@leplatrem I went through and made some changes based on your most recent comments.

I tried to add the suggested helper as an injected method in the bugzilla client returned from the get_bugzilla call, but something seemed off with that approach.

File: src/jbi/router.py [hunk 42]
```diff
@@ -1,34 +1,78 @@
...
+
+        action_module: ModuleType = importlib.import_module(current_action["action"])
+        callable_action = action_module.init(  # type: ignore
+            **current_action["parameters"]
+        )
+        content = callable_action(payload=request)
+        jbi_logger.info("request: %s, content: %s", request.json(), content)
+        return JSONResponse(content=content, status_code=200)
+    except IgnoreInvalidRequestError as exception:
+        jbi_logger.info("ignore-invalid-request: %s", exception, exc_info=True)
```
leplatrem, line 64: Are you sure this is going to be too much logging (and same level as executed actions)? 

The `exc_info` gives the stracktrace IIRC but we don't really care here IMO. 

I would suggest to either:
- create a dedicated logger `invalid_logger = logging.getLogging("ignored-requests")` so that we can filter it easily in prod
- switch to debug level here

src/jbi/router.py
```diff
@@ -1,34 +1,78 @@
...
+            raise IgnoreInvalidRequestError(
+                "whiteboard tag not found in configured actions"
+            )
+
+        action_module: ModuleType = importlib.import_module(current_action["action"])
+        callable_action = action_module.init(  # type: ignore
+            **current_action["parameters"]
+        )
+        content = callable_action(payload=request)
+        jbi_logger.info("request: %s, content: %s", request.json(), content)
```
leplatrem, line 61: nitpick: I would add one logging call before executing the action, so that we can catch issues before executing

leplatrem (approved):

src/jbi/router.py
bsieber-mozilla: Thanks for the callout!

src/jbi/router.py
bsieber-mozilla: Agreed, I've made the changes in a new PR as a fast follow; I'm going to resolve this and merge this in as-is for now.
https://github.com/mozilla/jira-bugzilla-integration/pull/34

PR #49, opened 2022-04-29
Reducing data displayed from visible projects insight

bsieber-mozilla:


src/jbi/services.py
```diff
@@ -59,4 +60,5 @@ def jbi_service_health_map():
 def jira_visible_projects():
     """Returns list of projects that are currently accessible"""
     jira = get_jira()
-    return jira.projects(included_archived=None)
+    visible_projects: List[Dict] = jira.projects(included_archived=None)
+    return [project["key"] for project in visible_projects]
```
leplatrem, line 64: Ideally there would be an associated test to be changed ;)

PR #52, opened 2022-04-30
Bugzilla EXPECTS a 200--else it's an unhashable response.

bsieber-mozilla:


src/jbi/router.py
```diff
@@ -40,6 +40,7 @@ def extract_current_action(  # pylint: disable=inconsistent-return-statements
 def execute_action(request: BugzillaWebhookRequest, action_map, settings):
     """Execute action"""
     try:
+        jbi_logger.info("request: %s", request.json())
```
leplatrem, line 43: I'm curious how this would look like in logs. 
You may have to use the `extra` keyword arg.
https://docs.python.org/3/library/logging.html#logging.debug
I realized that when working on https://github.com/leplatrem/logging-color-formatter/blob/189b21a7bbff3b54091c3d0994553082de2b5460/tests/test_formatter.py#L93-L94

PR #57, opened 2022-05-11
Adding remaining JBI V1 configuration.

bsieber-mozilla:


config/config.prod.yaml
```diff
@@ -31,6 +31,13 @@ actions:
     parameters:
       jira_project_key: FXATPS
       whiteboard_tag: fxatps
+  fxcm:
+    contact: tbd
+    description: tbd
```
leplatrem, line 36: The description may at least be the expended acronym?

leplatrem (approved):

config/config.prod.yaml
bsieber-mozilla: Yeah, not sure what it is--was pulling this from the firestore config; we'll have to reach out to Julien to go through the config together and update this file with contact and description info. 

I'll make it an issue to track here.

bsieber-mozilla: Looked it up on Jira--added it into the recent commit. I still think going over the config with stakeholders to polish it up would be worthwhile

PR #63, opened 2022-06-09
Allow actions to specify if they support private bugs.

Mossop:
* Adds an `allow_private` settings to actions, defaults to False.
* Adds parsing of bugzilla groups for actions to use.

By default this doesn't change existing actions but it allows actions to
opt in to syncing private bugs. A custom action my be written to only
sync depending on the groups a bug is a member of. This will only
support bugs in groups that the JBI bugzilla account is a member of.

Strictly speaking the changes to the default action are unnecessary but since any project wanting to support this would need them and we have to change internal code anyway it made sense to me to make them. If you want those removed let me know.

The only thing I've been unable to verify myself is the behaviour for a private bug that is member of a group that the JBI bugzilla API key is not a member of. It isn't clear if the webhook will just not deliver those to JBI in which case I would expect everything to be fine or if it will deliver something in which case JBI's bug lookup request would fail and throw an error. Not being able to create custom groups and additional users on the test bugzilla instance mean I couldn't check that.

I found it a little odd that Bugzilla's `bug.create` webhook for private bugs includes almost no information about the bug while the `comment.create` webhook for the same bug would include the full details!

### DIFF
--- a/src/jbi/bugzilla.py
+++ b/src/jbi/bugzilla.py
@@ -91,6 +91,7 @@ class BugzillaBug(BaseModel):
     whiteboard: Optional[str]
     keywords: Optional[List]
     flags: Optional[List]
+    groups: Optional[List]
     status: Optional[str]
     resolution: Optional[str]
     see_also: Optional[List]
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -16,6 +16,7 @@ class Action(YamlModel, extra=Extra.allow):
 
     action: str = "src.jbi.whiteboard_actions.default"
     enabled: bool = False
+    allow_private: bool = False
     parameters: dict = {}
 
     @root_validator
--- a/tests/unit/jbi/noop_action.py
+++ b/tests/unit/jbi/noop_action.py
@@ -8,18 +8,5 @@ A no-op action for testing.
 from src.jbi.bugzilla import BugzillaWebhookRequest
 
 
-def init(**kwargs):
-    """Function that takes required and optional params and returns a callable object"""
-    return NoopExecutor(**kwargs)
-
-
-class NoopExecutor:
-    """Callable class that encapsulates the no-op action."""
-
-    def __init__(self, **kwargs):
-        """Initialize DefaultExecutor Object"""
-        self.parameters = kwargs
-
-    def __call__(self, payload: BugzillaWebhookRequest):
-        """Called from BZ webhook when no-op action is used. Does nothing."""
-        pass
+def init(**parameters):
+    return lambda payload: {"parameters": parameters, "payload": payload}
--- a/tests/unit/jbi/noop_action.py
+++ b/tests/unit/jbi/noop_action.py
@@ -5,8 +5,6 @@ A no-op action for testing.
 `init` should return a __call__able
 """
 
-from src.jbi.bugzilla import BugzillaWebhookRequest
-
 
 def init(**parameters):
-    return lambda payload: {"parameters": parameters, "payload": payload}
+    return lambda payload: {"parameters": parameters, "payload": payload.json()}
--- a/tests/unit/jbi/noop_action.py
+++ b/tests/unit/jbi/noop_action.py
@@ -1,0 +1,25 @@
+"""
+A no-op action for testing.
+`init` is required.
+
+`init` should return a __call__able
+"""
+
+from src.jbi.bugzilla import BugzillaWebhookRequest
+
+
+def init(**kwargs):
+    """Function that takes required and optional params and returns a callable object"""
+    return NoopExecutor(**kwargs)
+
+
+class NoopExecutor:
+    """Callable class that encapsulates the no-op action."""
+
+    def __init__(self, **kwargs):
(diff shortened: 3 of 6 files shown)


src/jbi/bugzilla.py
```diff
@@ -91,6 +91,7 @@ class BugzillaBug(BaseModel):
     whiteboard: Optional[str]
     keywords: Optional[List]
     flags: Optional[List]
+    groups: Optional[List]
```
leplatrem, line 94: nitpick: we could leave for another pull-request if we don't really use them. Also, it's not harmful.

tests/unit/jbi/noop_action.py
```diff
@@ -0,0 +1,25 @@
...
+class NoopExecutor:
+    """Callable class that encapsulates the no-op action."""
+
+    def __init__(self, **kwargs):
+        """Initialize DefaultExecutor Object"""
+        self.parameters = kwargs
+
+    def __call__(self, payload: BugzillaWebhookRequest):
+        """Called from BZ webhook when no-op action is used. Does nothing."""
+        pass
```
leplatrem, line 25: nit: I wonder if we need all this scaffolding for a no-op action.

Since the action returned value is served as the JSON response, wouldn't something like this be enough for testing?

```py
def init(**parameters):
     return lambda payload: {"parameters": parameters, "payload": payload}
```

leplatrem (approved):
Thanks for your contribution!

> The only thing I've been unable to verify myself is the behaviour for a private bug that is member of a group that the JBI bugzilla API key is not a member of. 

Once this is verified, I believe having this option on the default action makes sense 👌 💯

tests/unit/jbi/noop_action.py
Mossop: Yeah that is simpler.

src/jbi/bugzilla.py
Mossop: True, do you want me to remove it?

Mossop:
I'm not sure why the github action is failing with my latest changes, it looks to be a networking problem.

tests/unit/jbi/test_router.py
```diff
@@ -10,31 +10,71 @@
...
+                    # https://fastapi.tiangolo.com/advanced/testing-events/
+
+                    response = anon_client.post(
+                        "/bugzilla_webhook", data=private_webhook_request_example.json()
+                    )
+                    assert response
+                    assert response.status_code == 200
+
+                    assert mocked_action.call_count == 1
+                    assert mocked_action.call_args.kwargs["payload"].bug.id == 654321
```
leplatrem, line 77: Alternatively, since the noop action returns the input parameters and payload, you can remove the mocking and read them from the JSON response

src/jbi/bugzilla.py
leplatrem: You can leave it, it just affects parsing

tests/unit/jbi/test_router.py
Mossop: Experimenting with this made me find a few issues with the action as-is.

Returning the payload from the action turns out to be a problem because we then attempt to JSON encode the payload and we fail with `Object of type BugzillaWebhookRequest is not JSON serializable`. I could convert it to a dict first but then we fail with `Object of type datetime is not JSON serializable`.

In the change I'm pushing I return the payload as a json string. That has the issue that we effectively have to double parse the response in order to get at it but it does work. Let me know if you'd prefer something else.

leplatrem: Yes, this is fine I think! 

It's good as it is, but if you want to remove the call to `parse_raw()`  in the test and have something equivalent to `.dict()`, you could try to return `{..., "payload": json.loads(payload.json())}` from the action. I'm not sure but dates should become strings in that case.

Mossop:
What are the next steps to getting this merged?

leplatrem:
> What are the next steps to getting this merged?

I'm waiting a quick confirmation from @bsieber-mozilla and @grahamalama that they're ok to merge this, and we'll roll it out

src/jbi/models.py
```diff
@@ -16,6 +16,7 @@ class Action(YamlModel, extra=Extra.allow):
 
     action: str = "src.jbi.whiteboard_actions.default"
     enabled: bool = False
+    allow_private: bool = False
```
bsieber-mozilla, line 19: *default = False ✅

leplatrem:
@Mossop As far as I understood, in order to deploy this for your project, the next steps are:
- add your entry in the config here https://github.com/mozilla/jira-bugzilla-integration/tree/main/config
- this will be deployed automatically on STAGE when merged
- test it :)
- this will be deployed automatically on PROD when tagged

Mossop:
Yes. I believe also that the Jira Integration Bot user in Bugzilla will need to be made a member of the group that we care about.

bsieber-mozilla:
@Mossop that is the right account; we might need dkl to provide the additional permissions. Can this be tested via non-prod before we rollout to production?
https://bugzilla-dev.allizom.org/user_profile?user_id=661118

PR #65, opened 2022-06-28
Remove unused mr2qa tag and add flowstate tag.

Mossop:
Sets up the flowstate tag allowing JBI to read from private bugs. We should verify that this is working correctly in staging first, I believe that https://bugzilla-dev.allizom.org/ already has the right pieces set up.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -24,6 +24,14 @@ actions:
     parameters:
       jira_project_key: FIDEFE
       whiteboard_tag: fidefe
+  flowstate:
+    contact: dtownsend@mozilla.com
+    description: Flowstate whiteboard tag
+    enabled: true
+    allow_private: true
+    parameters:
+      jira_project_key: MR2
+      whiteboard_tag: flowstate
   fxatps:
     contact: tbd
     description: Privacy & Security and Anti-Tracking Team whiteboard tag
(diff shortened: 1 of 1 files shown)


leplatrem:
@Mossop The docker image failed to build. I opened #66 to fix the issue. 

With regards to access to logs of deployment and execution, my knowledge is limited. @bsieber-mozilla can we document this somewhere?

leplatrem:
In order to see the config being deployed in STAGE, the nonprod config had to be modified too. See #68

Mossop:
I'm still not seeing this work but I just remembered when testing locally I also had to override the [dev jira server](https://github.com/mozilla/jira-bugzilla-integration/blob/main/src/app/environment.py#L25) to be https://mozilla-hub-sandbox-721.atlassian.net. I'm guessing that staging is still pointing to mozit-test which probably doesn't have the MR2 project defined in it.

leplatrem:
The action is configured and deployed https://stage.jbi.nonprod.cloudops.mozgcp.net/powered_by_jbi

I opened #71 to be able to know easily which Bugzilla and Jira are configured on the instance

leplatrem:
As now exposed on the stage server root URL, the configured servers are:

```
        "bugzilla_base_url": "https://bugzilla-dev.allizom.org",
        "jira_base_url": "https://mozit-test.atlassian.net/"
```

Mossop:
Ok so I don't think I have access to that Jira instance to know whether the MR2 project is defined there.

Mossop:
@leplatrem Can you give me some guidance on how I can verify this config?

PR #68, opened 2022-06-28
Add flowstate to nonprod

leplatrem:
If we want to test the `flowstate` use-case (private bugs) in DEV/STAGE, it has to be defined in the nonprod config.


bsieber-mozilla (approved):
LGTM

PR #69, opened 2022-06-29
Document how to inspect rollout

leplatrem:


bsieber-mozilla (approved):
Minor comment around the env's. There's no dev env at this time, just prod and nonprod; is this something we'd like to change?

PR #76, opened 2022-07-07
Refactor of API versus JBI, and leverage of pytest fixtures for smaller tests

leplatrem:
⚠️ Please review by commit ⚠️ 

**Rationale**

- The JBI domain lives in `src.jbi`
- The HTTP API can be treated as a View. View in `src.app` and Models and Controler are in `src.jbi`
- Contributors are more likely to contribute actions, and less API stuff. By separating the two, contributors only have to discover and learn `src.jbi`. 
- Keep `src.jbi` as focused as possible to JBI domain (no configuration parsing etc.)
- Unit tests of `src.jbi` do not involve API requests (less mocking, less verbose)
- Use our Pedantic models as much as possible, instead of `Dict`, to have high-level logics close to their definition, and have consistent types everywhere
- Use pytest fixtures to avoid repetitive declarations of payloads, which make the test verbose and hard context setup harder to understand 
- Also now much faster, `47 passed, 6 warnings in 0.27s`  versus `43 passed, 4 warnings in 5.28s`
- We should do this ASAP, so that contributors can mimic clean code instead of duplicating verbose code

### DIFF
--- a/src/app/api.py
+++ b/src/app/api.py
@@ -2,9 +2,9 @@
 Core FastAPI app (setup, middleware)
 """
 import logging
-import os
 import time
 from datetime import datetime
+from pathlib import Path
 from typing import Dict, List, Optional
 
 import sentry_sdk
@@ -24,10 +24,9 @@ from src.jbi.models import Actions
 from src.jbi.runner import IgnoreInvalidRequestError, execute_action
 from src.jbi.services import get_jira
 
-SRC_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+SRC_DIR = Path(__file__).parents[1]
 
-
-templates = Jinja2Templates(directory=os.path.join(SRC_DIR, "templates"))
+templates = Jinja2Templates(directory=SRC_DIR / "templates")
 
 settings = get_settings()
 
@@ -40,9 +39,7 @@ app = FastAPI(
 )
 
 app.include_router(monitor_router)
-app.mount(
-    "/static", StaticFiles(directory=os.path.join(SRC_DIR, "static")), name="static"
-)
+app.mount("/static", StaticFiles(directory=SRC_DIR / "static"), name="static")
 
 sentry_sdk.init(  # pylint: disable=abstract-class-instantiated  # noqa: E0110
     dsn=settings.sentry_dsn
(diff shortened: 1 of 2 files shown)


grahamalama (changes requested):
Going to do another deeper review, but I saw this 500 error off the bat.

leplatrem:
`Coverage report: 99%`

src/app/api.py
```diff
@@ -2,19 +2,32 @@
...
 from src.app.environment import get_settings
 from src.app.log import configure_logging
 from src.app.monitor import api_router as monitor_router
-from src.jbi.router import api_router as jbi_router
+from src.jbi.bugzilla import BugzillaWebhookRequest
+from src.jbi.models import Actions
+from src.jbi.runner import IgnoreInvalidRequestError, execute_action
+from src.jbi.services import get_jira
+
+SRC_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
grahamalama, line 27: It might me nice to switch to the `Pathlib` api here
```python
from pathlib import Path
# ...
SRC_DIR = Path(__file__).parents[1]
# ...
templates = Jinja2Templates(directory=SRC_DIR / "templates")
# ...
app.mount("/static", StaticFiles(directory=SRC_DIR / "static"), name="static")
```

src/jbi/runner.py
```diff
@@ -0,0 +1,84 @@
...
+    IGNORE = "ignore"
+    SUCCESS = "success"
+
+
+def execute_action(
+    request: BugzillaWebhookRequest,
+    actions: Actions,
+    settings: Settings,
+):
+    """Execute action"""
```
grahamalama, line 29: Not to address here, but can we add an issue to better document this function?

grahamalama (approved):
Some last optional comments, otherwise looks g2g

src/app/api.py
```diff
@@ -78,6 +92,56 @@ async def request_summary(request: Request, call_next):
...
+    enabled: Optional[bool] = None,
+    actions: Actions = Depends(configuration.get_actions),
+):
+    """API for `Powered By` endpoint"""
+    entries = actions.all()
+    context = {
+        "request": request,
+        "title": "Powered by JBI",
+        "num_configs": len(entries),
+        "data": {k: v.dict() for k, v in entries.items()},
```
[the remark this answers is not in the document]
leplatrem: I was thinking to convert objects to `dict()` but that may not be necessary indeed

PR #80, opened 2022-07-07
Move mypy to pyproject

leplatrem:


pyproject.toml
```diff
@@ -60,6 +60,10 @@ testpaths = [
 [tool.isort]
 profile = "black"
 
+[tool.mypy]
+python_version = "3.8"
```
grahamalama, line 64: Should this be "3.9" to match the Dockerfile?

grahamalama (approved):

PR #83, opened 2022-07-08
Only lookup a bug once through the API and correctly handle errors

Mossop:
Currently we retrieve the full bug in router and then again in the action. We should just do this once and pass it through to the action.


src/jbi/router.py
```diff
@@ -66,7 +66,14 @@ def execute_action(
...
             raise IgnoreInvalidRequestError("no bug data received")
 
-        bug_obj = getbug_as_bugzilla_object(request=request)
+        try:
+            bug_obj = getbug_as_bugzilla_object(request=request)
+        except Exception as ex:
+            logger.exception("Failed to get bug: %s", ex, extra=log_context)
+            raise IgnoreInvalidRequestError(
+                "bug not accessible or bugzilla down"
+            ) from ex
```
leplatrem, line 75: 👍

src/jbi/whiteboard_actions/default.py
```diff
@@ -36,15 +36,13 @@ def __init__(self, whiteboard_tag, jira_project_key, **kwargs):
         self.jira_client = get_jira()
 
     def __call__(  # pylint: disable=inconsistent-return-statements
-        self, payload: BugzillaWebhookRequest
+        self, payload: BugzillaWebhookRequest, bug_obj: BugzillaBug
```
leplatrem, line 39: I'm not sure whether it is cleaner than your proposal, but alternatively, to avoid changing actions signatures, we could add a property to the `BugzillaWebhookRequest` class:

```py
@functools.cached_property
  def bugzilla_object(self):
    if not self.bug:
         raise ValueError("boom")
    return getbug_as_bugzilla_object(self)
```

Mossop: That's a good idea, I'll do that.

PR #84, opened 2022-07-08
Remove top-level `actions` config key

grahamalama:
I noticed that we had a top-level actions key in project config. This seemed somewhat unnecessary since there aren't any other config objects besides actions. That meant that any time we wanted to access actions, we'd need to do `actions_instance.actions.<whatever>`.

This PR removes the top-level `actions` key in config and modifies the `Actions` Pydantic model to give it a [`__root__`](https://pydantic-docs.helpmanual.io/usage/models/#custom-root-types) property to contain the actions.

### DIFF
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -16,6 +16,7 @@ class Action(YamlModel):
     Action is the inner model for each action in the configuration file"""
 
     action: str = "src.jbi.whiteboard_actions.default"
+    contact: str
     description: str
     enabled: bool = False
     allow_private: bool = False
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -16,8 +16,8 @@ class Action(YamlModel):
     Action is the inner model for each action in the configuration file"""
 
     action: str = "src.jbi.whiteboard_actions.default"
-    enabled: bool = False
     description: str
+    enabled: bool = False
     allow_private: bool = False
     parameters: dict = {}
 
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -17,6 +17,7 @@ class Action(YamlModel):
 
     action: str = "src.jbi.whiteboard_actions.default"
     enabled: bool = False
+    description: str
     allow_private: bool = False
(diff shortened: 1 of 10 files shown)


src/jbi/models.py
```diff
@@ -17,6 +17,7 @@ class Action(YamlModel):
 
     action: str = "src.jbi.whiteboard_actions.default"
     enabled: bool = False
+    description: str
     allow_private: bool = False
     parameters: dict = {}
```
bsieber-mozilla, line 22:
```suggestion
    action: str = "src.jbi.whiteboard_actions.default"
    allow_private: bool = False    
    contact: str
    description: str
    enabled: bool = False
    parameters: dict = {}
```

bsieber-mozilla: Suggestion: order params as expected in config; add `contact` field.

bsieber-mozilla (approved):
Minor suggestion; overall LGTM

src/jbi/models.py
grahamalama: https://github.com/mozilla/jira-bugzilla-integration/pull/84/commits/f0d07aca4bba3b8a97f3156892a88b2985d039d6

src/jbi/models.py
```diff
@@ -57,22 +59,24 @@ class Config:
...
+        return self.__root__[item]
 
     def get(self, tag: Optional[str]) -> Optional[Action]:
         """Lookup actions by whiteboard tag"""
-        return self.actions.get(tag.lower()) if tag else None
-
-    def all(self):
-        """Return mapping of all actions"""
-        return self.actions
+        return self.__root__.get(tag.lower()) if tag else None
```
leplatrem, line 75: Shouldn't we move this to `__getitem__()` ?

leplatrem (approved):

src/jbi/models.py
leplatrem: Yeah, but one could rely on the other to match the `actions[item.lower()]`  and `actions[None]` behaviour

PR #86, opened 2022-07-11
Use orjson for serialising responses.

Mossop:
When creating a bug we embed the response of the bugzilla update_bugs
call into the repsonse. This contains a datetime instance which is not
serialisable by the default JSONResponse and so currently JBI throws
an internal server error back to Bugzilla which then repeats the
request a second time which is effectively ignored because by then the
bug has the see_also field set.

This switches to ORJSONResponse which can serialise datetime and a
few other types correctly.


leplatrem (approved):

grahamalama:
Does https://fastapi.tiangolo.com/tutorial/encoder/ do this for us?

PR #87, opened 2022-07-11
Switch Flowstate to the default_with_assignee_and_status action

Mossop:


leplatrem:
@Mossop Don't you want it in `nonprod` config too/instead?

Mossop:
> @Mossop Don't you want it in `nonprod` config too/instead?

Ah yes!

leplatrem (approved):

PR #90, opened 2022-07-11
Constrain `Action.contact` field to email address(es)

grahamalama:
Constrain `contact` field so that it only accepts an email address or list of email addresses.

For now, it also accepts a "tbd" string literal (since many actions have that as their contact at the moment), but eventually we'll want to remove that option.

### DIFF
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -17,7 +17,7 @@ class Action(YamlModel):
     Action is the inner model for each action in the configuration file"""
 
     action: str = "src.jbi.whiteboard_actions.default"
-    # TODO: Remove the tbd literal option when all actions have contact info
+    # TODO: Remove the tbd literal option when all actions have contact info # pylint: disable=fixme
     contact: Union[EmailStr, List[EmailStr], Literal["tbd"]]
     description: str
     enabled: bool = False
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -3,11 +3,12 @@ Python Module for Pydantic Models and validation
 """
 import functools
 import importlib
+import warnings
 from inspect import signature
 from types import ModuleType
-from typing import Any, Callable, Dict, Mapping, Optional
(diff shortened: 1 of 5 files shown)


src/jbi/models.py
```diff
@@ -75,16 +77,21 @@ def get(self, tag: Optional[str]) -> Optional[Action]:
...
+           whiteboard tag parameter)
+         - If the action's contact is "tbd", emit a warning.
         """
         if not actions:
             raise ValueError("no actions configured")
         for name, action in actions.items():
             if name.lower() != action.parameters["whiteboard_tag"]:
                 raise ValueError("action name must match whiteboard tag")
+            if action.contact == "tbd":
+                warnings.warn(f"Provide contact data for `{name}` action.")
```
bsieber-mozilla, line 95: 👍

bsieber-mozilla (approved):
Thanks for these changes!

LGTM 🚢

src/jbi/models.py
```diff
@@ -16,7 +17,8 @@ class Action(YamlModel):
     Action is the inner model for each action in the configuration file"""
 
     action: str = "src.jbi.whiteboard_actions.default"
-    contact: str
+    # TODO: Remove the tbd literal option when all actions have contact info
```
bsieber-mozilla, line 20: #28 [lint 5/6] RUN pylint src tests
#28 5.251 ************* Module src.jbi.models
#28 5.251 src/jbi/models.py:20:5: W0511: TODO: Remove the tbd literal option when all actions have contact info (fixme)

bsieber-mozilla, line 20:
```suggestion
    # TODO: Remove the tbd literal option when all actions have contact info # pylint: disable=fixme
```

I think this could solve the error? Unless we want a global change to allow ToDo's

grahamalama: Resolved with your suggested commit below. Thanks!

PR #92, opened 2022-07-12
Use pytest-dotenv

grahamalama:
This PR adds [pytest-dotenv](https://pypi.org/project/pytest-dotenv/) and uses it to configure environment variables for testing. This means that we don't need to remember to set them in places like Docker containers, Github actions, etc.

It also renames the action workflow files for linting and testing, since they don't use the Docker container anymore as of b8dc0b51a488e236a30c5f43dd9e9054eade79d5


bsieber-mozilla (approved):
LGTM. Thanks for this consolidation for the env vars!

PR #88, opened 2022-07-11
Align test & lint

grahamalama:
Chatted with @leplatrem about #81, and I added some more changes on top of that PR here.

The idea with this PR is that `infra/lint.sh` and `infra/test.sh` are the "source of truth" for how we want to run linting and tests respectively. We then use these scripts in precommit, make, Github Actions, and wherever else we want to run tests and linting.

For Github Actions, we also now run linting and tests directly in the action runner, rather than running through Docker. This resulted in a speed-up of the checks.

Fixes #79 
Fixes #67

PR #102, opened 2022-07-13
Test build on pull requests

leplatrem:
Since #88 we don't build the container images anymore. 

The problem with this, is that we discover container issues only when we push tags.

This pull-request is an attempt to improve that, by building the container on each pull-request. 
I tried to reuse the existing Github Action, and executing the `publish` step only on main branch and tags. I'm not sure how to test that though.


bsieber-mozilla (approved):
LGTM

PR #109, opened 2022-07-13
Install latest Poetry in container

leplatrem:
This removes the warning about old version being installed.

It has the downside of potentially not being fully reproducible, but I think that this is acceptable if we have #102

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,10 +39,9 @@ RUN apt-get install --no-install-recommends -y \
       curl \
       build-essential
 
-# Install Poetry - respects $POETRY_VERSION & $POETRY_HOME
+# Install Poetry - respects $POETRY_HOME
 USER app
-ENV POETRY_VERSION=1.1.5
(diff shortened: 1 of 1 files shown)


leplatrem:
I also ran into this:

```
[2/3] STEP 8/8: RUN poetry install --no-dev --no-root
Traceback (most recent call last):
  File "/opt/poetry/bin/poetry", line 17, in <module>
    from poetry.console import main
  File "/opt/poetry/lib/poetry/console/__init__.py", line 1, in <module>
    from .application import Application
  File "/opt/poetry/lib/poetry/console/application.py", line 3, in <module>
    from cleo import Application as BaseApplication
ModuleNotFoundError: No module named 'cleo'
```

grahamalama:
@leplatrem I just noticed [this](https://python-poetry.org/docs/master/#installing-with-the-official-installer) in the docs -- looks like a new method to install poetry. Just got a working build with it.

```
curl -sSL https://install.python-poetry.org | python3 -
```

grahamalama:
I'm curious why this PR triggered a cloudops "deployment". I don't think it deployed anything -- just built the container as a test?

Issue #107, opened 2022-07-13
Container not built nor ran with current host user id


leplatrem:
`make test` fails with:

```
Creating jira-bugzilla-integration_web_run ... done
/docker-entrypoint.sh: 15: exec: bin/test.sh: Permission denied
ERROR: 126
make: *** [test] Error 126
```

The current user id is not passed as build arg (`id -u`), and the host user id does not match the container `app` id (10001).

leplatrem:
I have tried several thing, without success, like:

```diff
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -4,6 +4,10 @@ services:
     build:
       context: .
       target: development
+      args:
+        userid: ${_UID:-10001}
+        groupid: ${_GID:-10001}
+    user: ${_UID:-10001}:${_GID:-10001}
     volumes:
       - .:/app
     ports:
```

and executing the build with ``_UID=`id -u` docker-compose build``, but the container is still built with `10001:10001`

bsieber-mozilla:
I'm not seeing this issue locally; going to close it since there was a PR merge.

Feel free to reopen if regressed or still occurring

PR #111, opened 2022-07-14
Remove `builder-base` and `development` image stages, run linting and tests outside Docker

### LINKED ISSUE #110: Remove `development` stage from Dockerfile
Since our development environment is relatively easy to set up to run tests & linting, and those things are not run through Docker in CI anymore (since #88), we should remove the `development` stage of the Dockerfile. We should still configure a way to run the production container in "development" mode through environment variables and/or `uvicorn.run()` params.

grahamalama:
In this PR, we:
- remove the `development` and `builder-base` stages of the Dockerfile so all that's left is a `base` stage for downloading Python dependencies and a `production` stage for running the application code. We still run the image in a "development" mode locally through the use of environment variables and overriding the image's `CMD` instruction.
- run linting and testing locally (rather than through a Docker image)
- add a `.dockerignore` so that only application code ends up in the production image

Closes #110


Dockerfile
```diff
@@ -1,101 +1,46 @@
...
+FROM python:3.10.5 as base
+ENV PIP_NO_CACHE_DIR=off \
     PIP_DEFAULT_TIMEOUT=100 \
+    PIP_DISABLE_PIP_VERSION_CHECK=on \
     POETRY_HOME="/opt/poetry" \
-    POETRY_VIRTUALENVS_IN_PROJECT=true \
     POETRY_NO_INTERACTION=1 \
-    PYSETUP_PATH="/opt/pysetup" \
-    VENV_PATH="/opt/pysetup/.venv" \
-    PORT=8000
```
leplatrem, line 14: This is necessary I think.

I ran `make shell`  from a fresh environment, and got this:

```
[2/2] STEP 11/13: EXPOSE $PORT
ERROR: Service 'web' failed to build : error building at STEP "EXPOSE $PORT": EXPOSE requires at least one argument

make: *** [shell] Error 1
```

config/local_dev.env
```diff
@@ -1,5 +1,7 @@
 # Environment variables for the local development environment
 
+ENV=nonprod
```
leplatrem, line 3: Should it be `local`?

I guess it would imply having a `config.local.yaml`. But it would be better than having the `devtest` action used in unit tests in `config.nonprod.yaml`

leplatrem (commented):
Wao 🤯 big shrink! 

Weird, the missing `PORT` var is blocking the build on my env, and it wasn't caught in the Github Action build...

Should we have a Github Action that does a `make start`  and `curl --connect-timeout 30 --retry 300  --retry-delay 5 http://localhost:8000/` to make sure we don't break anything in docker-compose?


And I'm also facing #107, but I guess you don't, and it can be fixed in another PR.
```
$ make start
docker-compose up
/opt/homebrew/lib/python3.9/site-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated
  "class": algorithms.Blowfish,
Creating jira-bugzilla-integration_web_1 ... done
Attaching to jira-bugzilla-integration_web_1
web_1  | ERROR (catatonit:2): failed to exec pid1: Permission denied
jira-bugzilla-integration_web_1 exited with code 1
```

Random driveby: what would you think about prefixing makefile targets that go through Docker with `container-` or `docker-`?
eg.  `make start` vs `make docker-start`

Makefile
```diff
@@ -20,29 +24,33 @@ help:
 	@echo ""
 	@echo "  help    - see this text"
 
+install: $(INSTALL_STAMP)
+$(INSTALL_STAMP): poetry.lock
+	poetry install
```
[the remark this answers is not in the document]
grahamalama: https://github.com/mozilla/jira-bugzilla-integration/pull/111/commits/328b184bfdf1c4b0c0ebca483d18e0e28242f607

leplatrem (approved):

leplatrem:
> I'm https://github.com/mozilla/jira-bugzilla-integration/pull/109#issuecomment-1184629298 not really sure what [this](https://github.com/mozilla/jira-bugzilla-integration/blob/07f2f3d3fcefd1ffc965e300ea44ae117e8fa545/.github/workflows/build-publish.yaml#L37-L38) means -- I think it's because we have this configured in our Github actions? See [this deployment / environment documentation](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#using-an-environment).

I missed your previous comment. 

I looks like we don't use these features. Even if that does not seem to affect anything, I would be in favor of removing anything that we don't use, and keep only the strict necessary and sufficient stuff

config/local_dev.env
bsieber-mozilla: Alternatively, we opt for one config and make the chosen environments a configuration parameter.

Then the local/nonprod/prod actions would be in a consolidated yaml but enabled/disabled through whichever parameter. (Boolean flag per environment, list of all environments, etc?)

PR #112, opened 2022-07-15
Tidy up FastAPI route functions

grahamalama:
In this PR, we:
* return objects directly instead of wrapping in response classes
FastAPI will automatically encode objects as JSON when they are returned
directly ([docs](https://fastapi.tiangolo.com/advanced/response-directly/?h=jsonable#return-a-response-directly))*

* use multiple route decorators for monitor functions with the same response
* remove extraneous route function parameters

*@Mossop, I believe this also means we can remove `orjson` (and revert #86)


leplatrem (approved):

bsieber-mozilla (approved):
LGTM! Nice tidying

PR #118, opened 2022-07-18
Refactor Docker build and push jobs

grahamalama:
- remove login for build step since it was blocking CI passing for dependabot
  PRs
- remove `cloudops` key environment since we don't use Actions deployment
  environments
- remove `medatadata-action` since we weren't using any of its output
- save built images with a filepath that includes the commit SHA to
  prevent overwriting
- save containers for 7 days to prevent the container being deleted
  before a PR is merged

### DIFF
--- a/.github/workflows/build-publish.yaml
+++ b/.github/workflows/build-publish.yaml
@@ -57,12 +57,12 @@ jobs:
 
       - name: Save image as artifact
         run: |
-          docker save -o /tmp/container-$IMAGE_TAG.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
+          docker save -o /tmp/container-$GITHUB_HEAD_REF.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
 
       - uses: actions/upload-artifact@v3
         with:
           name: container-image
-          path: /tmp/container-${{ env.IMAGE_TAG }}.tar.gz
+          path: /tmp/container-${{ env.GITHUB_HEAD_REF }}.tar.gz
           retention-days: 7 # Instead of default 90 days
 
   publish:
@@ -78,11 +78,11 @@ jobs:
       - uses: actions/download-artifact@v3
         with:
           name: container-image
-          path: /tmp/container-${{ env.IMAGE_TAG }}.tar.gz
+          path: /tmp/container-${{ env.GITHUB_HEAD_REF }}.tar.gz
 
       - name: Load container from artifact
         run: |
-          docker load -i /tmp/container-$IMAGE_TAG.tar.gz
+          docker load -i /tmp/container-$GITHUB_HEAD_REF.tar.gz
 
       - name: Login to Docker Hub
         uses: docker/login-action@v1
--- a/.github/workflows/build-publish.yaml
+++ b/.github/workflows/build-publish.yaml
@@ -62,7 +62,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           name: container-image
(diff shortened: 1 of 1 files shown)


.github/workflows/build-publish.yaml
```diff
@@ -72,33 +56,33 @@ jobs:
...
-          docker save -o /tmp/container.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
+          docker save -o /tmp/container-$IMAGE_TAG.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
 
       - uses: actions/upload-artifact@v3
         with:
           name: container-image
-          path: /tmp/container.tar.gz
-          retention-days: 1  # Instead of default 90 days
+          path: /tmp/container-$IMAGE_TAG.tar.gz
+          retention-days: 7 # Instead of default 90 days
```
grahamalama, line 66: Allued to this in the PR description, but my thinking here was:
- author submits PR, image gets built
- PR is reviewed and approved 25 hours later

I think the built image would no longer be available and the push job would fail.

This is just a larger "magic number" further in the future, but hopefully it gives us more time for the review + merge cycle.

.github/workflows/build-publish.yaml
```diff
@@ -34,34 +34,18 @@ jobs:
   build:
     runs-on: ubuntu-latest
     needs: prepare
-    environment:
-        name: cloudops
```
grahamalama, line 38: We will potentially need to move the `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets from Environment Secrets to Repository Secrets for publishes to work.

.github/workflows/build-publish.yaml
```diff
@@ -72,43 +56,41 @@ jobs:
           docker build --file Dockerfile -t $GITHUB_REPOSITORY:$IMAGE_TAG .
 
       - name: Save image as artifact
-        env:
-          IMAGE_TAG: ${{ needs.prepare.outputs.FULL_IMAGE_TAG }}
         run: |
-          docker save -o /tmp/container.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
+          docker save -o /tmp/container-$IMAGE_TAG.tar.gz $GITHUB_REPOSITORY:$IMAGE_TAG
```
grahamalama, line 60: This might actually save a lot of container images as authors add commits to a PR. Should we maybe use `GITHUB_REF_NAME` name instead?

.github/workflows/build-publish.yaml
bsieber-mozilla: I think that might require some assistance from @kkleemola

.github/workflows/build-publish.yaml
```diff
@@ -3,13 +3,13 @@ name: Build and Publish Docker image
 on:
   pull_request:
     branches:
-    - main
+      - main
```
bsieber-mozilla, line 6: yamllint?

.github/workflows/build-publish.yaml
grahamalama: > I think that might require some assistance from @kkleemola

We were chatting about this a bit this morning, but I wanted to get input from you (@bsieber-mozilla) and @leplatrem before we moved forward with this

.github/workflows/build-publish.yaml
grahamalama: Ahh no -- I think the Prettier extension I have installed in VSCode

grahamalama:
Going to close this in favor of #120

PR #120, opened 2022-07-18
Rewrite build-publish workflow to use mainly Docker actions

grahamalama:
This PR rewrites the `build-publish` workflow to mainly use:
- https://github.com/marketplace/actions/docker-metadata-action
- https://github.com/marketplace/actions/build-and-push-docker-images

As a result of this PR, the following should be true:
- the image is built on every PR (as a CI check)
- the image is published in this way:

| Event             | Tags                        |
|-------------------|-----------------------------|
| push tag vX.Y.Z   | `latest`, `X.Y.Z`           |
| PR merged to main | `latest`, `sha-<short sha>` |

### DIFF
--- a/.github/workflows/build-publish.yaml
+++ b/.github/workflows/build-publish.yaml
@@ -3,113 +3,49 @@ name: Build and Publish Docker image
 on:
   pull_request:
     branches:
-    - main
-
+      - main
   push:
     branches:
-    - main
+      - main
     tags:
-    - v[0-9]+.[0-9]+.[0-9]+
+      - v[0-9]+.[0-9]+.[0-9]+
 
 jobs:
-  prepare:
+  build-and-publish:
     runs-on: ubuntu-latest
-    outputs:
-      FULL_IMAGE_TAG: ${{ steps.tag.outputs.tag }}
     steps:
-      - name: Set Tag
-        id: tag
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Build `version.json` file
         run: |
-          export CI_COMMIT_SHORT_SHA=$(echo $GITHUB_SHA | cut -c1-8)
-          echo $CI_COMMIT_SHORT_SHA;
-          echo ${GITHUB_REF##*/};
-          echo $version_pattern;
-          if [[ ${GITHUB_REF} =~ $version_pattern ]]; then
-            echo "::set-output name=tag::${GITHUB_REF##*/}"
-          else
-            echo "::set-output name=tag::${GITHUB_REF##*/}-$CI_COMMIT_SHORT_SHA"
-          fi
-        env:
-          version_pattern: "tags\\/v[0-9]+\\.[0-9]+\\.[0-9]+"
-  build:
-    runs-on: ubuntu-latest
-    needs: prepare
-    environment:
-        name: cloudops
-
-    steps:
-      - name: Docker image tag
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-publish.yaml
```diff
@@ -3,113 +3,49 @@ name: Build and Publish Docker image
...
-
-      - name: Build Docker image
-        env:
-          IMAGE_TAG: ${{ needs.prepare.outputs.FULL_IMAGE_TAG }}
-        run: |
-          printf '{\n    "commit": "%s",\n    "version": "%s",\n    "image_tag": "%s",\n    "source": "%s",\n    "build": "%s"\n}\n' \
+          printf '{\n    "commit": "%s",\n    "version": "%s",\n    "source": "%s",\n    "build": "%s"\n}\n' \
             "$GITHUB_SHA" \
             "$GITHUB_REF" \
-            "$IMAGE_TAG" \
```
grahamalama, line 69: I think we're safe to remove this since it's not a standard `version.json` field and the SHA is captured by the `commit` field. Happy to put in back though.

.github/workflows/build-publish.yaml
```diff
@@ -3,113 +3,49 @@ name: Build and Publish Docker image
...
-
-      - name: Load container from artifact
-        run: |
-          docker load -i /tmp/container.tar.gz
+          images: $GITHUB_REPOSITORY
+          # https://github.com/marketplace/actions/docker-metadata-action#tags-input
+          tags: |
+            type=semver,pattern={{version}}
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=sha,enable={{is_default_branch}}
```
grahamalama, line 37: You'll see in the first action run for this PR that there were a warnings:
```
Warning: No Docker image version has been generated. Check tags input.
Warning: No Docker tag has been generated. Check tags input.
```

I'm not _exactly_ sure about this, but my thinking is that since none of these rules apply (this wasn't a push to `main` nor a tagged commit), tags and and a version weren't generated. Maybe we should remove the `enabled={{is_default_branch}}` and `github.event_name != 'pull_request'` for a run or two to test that this works? Or YOLO and see what happens after a merge

.github/workflows/build-publish.yaml
```diff
@@ -3,113 +3,49 @@ name: Build and Publish Docker image
...
-          else
-            echo "::set-output name=tag::${GITHUB_REF##*/}-$CI_COMMIT_SHORT_SHA"
-          fi
-        env:
-          version_pattern: "tags\\/v[0-9]+\\.[0-9]+\\.[0-9]+"
-  build:
-    runs-on: ubuntu-latest
-    needs: prepare
-    environment:
-        name: cloudops
```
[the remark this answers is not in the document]
leplatrem: Yes, if not used

.github/workflows/build-publish.yaml
```diff
@@ -3,113 +3,49 @@ name: Build and Publish Docker image
...
+          images: $GITHUB_REPOSITORY
+          # https://github.com/marketplace/actions/docker-metadata-action#tags-input
+          tags: |
+            type=semver,pattern={{version}}
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=sha,enable={{is_default_branch}}
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v1
+        if: github.event_name != 'pull_request'
```
leplatrem, line 40: Without this condition, we could test that our configured credentials remain valid :)

leplatrem (approved):
Super clean 💯

PR #123, opened 2022-07-19
Bump deploy ev

bsieber-mozilla:
Tagging minor version. (Will trigger deploy.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "jira-bugzilla-integration"
-version = "2.0.1"
+version = "2.0.2"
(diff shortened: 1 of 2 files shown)


bsieber-mozilla:
Might close--attempting rebuild through gh actions

grahamalama (commented):
- If this is a minor version bump, should we be making the version string `2.1.0`?
- Can we update all of the `2.0.1` values in the README?
- Has [enough happened](https://github.com/mozilla/jira-bugzilla-integration/compare/v2.0.1...1182a65296d3a64c2693fc67966320315f1bca66/) since the last release where this is potentially a major version change? (I'm not sure about this one -- what do you think?)

PR #124, opened 2022-07-19
Pushing patch version

bsieber-mozilla:


bsieber-mozilla:
https://github.com/mozilla/jira-bugzilla-integration/pull/123#pullrequestreview-1043958497

bsieber-mozilla:
> If this is a minor version bump, should we be making the version string 2.1.0?
Meant to be a patch. Just to unblock stakeholders.

> Can we update all of the 2.0.1 values in the README?
Yes.

> Has [enough happened](https://github.com/mozilla/jira-bugzilla-integration/compare/v2.0.1...1182a65296d3a64c2693fc67966320315f1bca66/) since the last release where this is potentially a major version change? (I'm not sure about this one -- what do you think?)

Worth a larger discussion--but I can see a v3.0.0 since the API/config has changed a lot; I'd be up for changing it in an alternate PR though

PR #125, opened 2022-07-20
Change default log level to `debug`

grahamalama:


leplatrem (approved):
We are changing this because:
- we want debug messages both in local dev and stage
- we can always set log level to INFO in prod if logs get too verbose (after stabilization phase)

PR #127, opened 2022-07-20
Check visible projects in heartbeat (fixes #126)

### LINKED ISSUE #126: Check that configured projects exist in heartbeat endpoint
We are currently in a situation in STAGE where some configured projects are not visible to the Jira client.

I think it's reasonable to receive an alert as soon as possible in that case.

We could have a startup check, but it would be safer to include it in the heartbeat, so that we receive an alert if a project is deleted (without JBI restarting)

leplatrem:
Fixes #126

### DIFF
--- a/src/jbi/services.py
+++ b/src/jbi/services.py
@@ -51,26 +51,23 @@ def _jira_check_health(actions: Actions) -> ServiceHealth:
     """Check health for Jira Service"""
     jira = get_jira()
     server_info = jira.get_server_info(True)
+    is_up = server_info is not None
     health: ServiceHealth = {
-        "up": server_info is not None,
+        "up": is_up,
+        "all_projects_are_visible": is_up and _all_jira_projects_visible(jira, actions),
     }
-    if server_info is None:
-        return health
+    return health
 
+
+def _all_jira_projects_visible(jira, actions: Actions) -> bool:
     visible_projects = {project["key"] for project in jira_visible_projects(jira)}
-    configured_projects = {
-        action.parameters["jira_project_key"]
-        for action in actions.__root__
-        if "jira_project_key" in action.parameters
-    }
-    missing_projects = configured_projects - visible_projects
+    missing_projects = actions.configured_jira_projects_keys - visible_projects
     if missing_projects:
         logger.error(
-            "Projects %s are not visible with configured credentials", missing_projects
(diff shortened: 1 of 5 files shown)


src/jbi/services.py
```diff
@@ -36,17 +47,35 @@ def bugzilla_check_health() -> ServiceHealth:
...
     server_info = jira.get_server_info(True)
-    health: ServiceHealth = {"up": server_info is not None}
+    health: ServiceHealth = {
+        "up": server_info is not None,
+    }
+    if server_info is None:
+        return health
+
+    visible_projects = {project["key"] for project in jira_visible_projects(jira)}
+    configured_projects = {
```
grahamalama, line 61: This _almost_ feels like something that should be added to `Actions`, maybe as a property:
```python
@functools.cached_property
def configured_jira_projects(self) -> Set[str]:
    return {
        action.parameters["jira_project_key"]
        for action in self.__root__
        if "jira_project_key" in action.parameters
    }
```

src/jbi/services.py
```diff
@@ -36,17 +47,35 @@ def bugzilla_check_health() -> ServiceHealth:
...
+        for action in actions.__root__
+        if "jira_project_key" in action.parameters
+    }
+    missing_projects = configured_projects - visible_projects
+    if missing_projects:
+        logger.error(
+            "Projects %s are not visible with configured credentials", missing_projects
+        )
+
+    health["all_projects_are_visible"] = not missing_projects
```
grahamalama, line 72: It _almost_ feels like we should abstract the above into a separate function, like:
```python
health["all_projects_are_visible"] = jira_configured_projects_available(actions)
```
where `jira_configured_projects_available` is:
```python
def jira_configured_projects_available(actions: Actions) -> bool:
    jira = get_jira()
    visible_projects = {project["key"] for project in jira_visible_projects(jira)}
    missing_projects = actions.configured_jira_projects - visible_projects
    if missing_projects:
        logger.error(
            "Projects %s are not visible with configured credentials",
            missing_projects,
        )
    return not missing_projects
```

I don't feel particularly strongly about this, just trying things out.

grahamalama (approved):
Looks good. Trying a few things out, but also ready to 🚢 as is

src/jbi/services.py
leplatrem: I like it ;)

PR #130, opened 2022-07-21
Fix make from fresh clone

leplatrem:
Without this, I would get:

```
touch .venv/.install.stamp
touch: .venv/.install.stamp: No such file or directory
make: *** [.venv/.install.stamp] Error 1
```

Because Poetry wouldn't create the `.venv` folder


Makefile
```diff
@@ -6,7 +6,6 @@ _GID ?= 10001
 
 VENV := $(shell echo $${VIRTUAL_ENV-.venv})
 INSTALL_STAMP = $(VENV)/.install.stamp
-POETRY_VIRTUALENVS_IN_PROJECT = true
```
grahamalama, line 9: Ahh, so this wasn't doing what I thought it was doing 😅

leplatrem: 😊 indeed

grahamalama (approved):

PR #131, opened 2022-07-21
 Read settings from .env file (fixes #53) 

### LINKED ISSUE #53: Document application settings from ENV vars
There should be a section in README to explicit how to set the application settings

leplatrem:
Without this, I would not be able to run the app with `make start` on a fresh clone


Makefile
```diff
@@ -48,15 +49,18 @@ lint: $(INSTALL_STAMP)
 	bin/lint.sh
 
 .PHONY: start
-start:
+start: $(INSTALL_STAMP) $(DOTENV_FILE)
 	poetry run python -m src.app.api
 
+$(DOTENV_FILE):
+	cp -n config/local_dev.env $(DOTENV_FILE)
```
grahamalama, line 56: maybe we should move this `local_dev.env` file to the root level and use the `.env.example` convention to make what's happening here a little more obvious.

grahamalama (approved):

grahamalama:
From the action run:
> src/app/environment.py:39:4: C0115: Missing class docstring (missing-class-docstring)

Maybe our bar for linting is a little high. We should still emit these warnings because it's usually good practice to have docstrings, but in this case the class in question doesn't really _need_ a doc string.

Maybe we should only fail pylint on `Warning` and above 🤔

leplatrem:
> Maybe our bar for linting is a little high.

We already had a similar docstring in `models.py`, it's ok to keep it I think

PR #134, opened 2022-07-21
Add prometheus metrics (fixes #62, fixes #23)

### LINKED ISSUE #23: Add StatsD (or Prometheus) metrics
- Count accepted / rejected BZ events
- Measure API calls latency
- Measure execution time

Configure client from `STATSD_DSN` env var. Skip if empty.

### LINKED ISSUE #62: Usage metrics for JBI
Context:  Feedback from the ITEO program review 

* metrics for each configured action about how many syncs it sent from BZ to Jira 
* can go to grafana
* gives us an idea of how much manual work JBI has saved / automated

leplatrem:
Fixes #62 
Fixes #23

### DIFF
--- a/src/jbi/services.py
+++ b/src/jbi/services.py
@@ -1,9 +1,11 @@
 """Services and functions that can be used to create custom actions"""
+import contextlib
 import logging
 from typing import Dict, List
 
 import bugzilla as rh_bugzilla
 from atlassian import Jira
+from prometheus_client import Counter, Summary
 
 from src.app import environment
 from src.jbi.models import Actions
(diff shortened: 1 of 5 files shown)


src/jbi/services.py
```diff
@@ -13,17 +14,45 @@
...
+    the specieds methods are called.
+    """
+
+    counters: Dict[str, Counter] = {}
+
+    def __init__(self, wrapped, methods):
+        self.wrapped = wrapped
+        for method in methods:
+            if method not in self.counters:
+                self.counters[method] = Counter(method, f"Call of {method}()")
```
leplatrem, line 28: maybe there should be a prefix string here (eg. `prefix="jira."`)

bsieber-mozilla: or have `prefix` be a param; so this could be reused for the other services?

bsieber-mozilla (approved):
LGTM! thanks!

bsieber-mozilla:
Where will we be able to view these metrics?

leplatrem:
> Where will we be able to view these metrics?
>
I was going to ask you. Since you don't know, I guess that it was never setup. I assume you did since there were mentions of Prometheus in the Gunicorn conf. 

That means I need to make sure the Promotheus client is actually sending these metrics before merging this PR

bsieber-mozilla:
> > Where will we be able to view these metrics?
> 
> I was going to ask you. Since you don't know, I guess that it was never setup. I assume you did since there were mentions of Prometheus in the Gunicorn conf.
> 
> That means I need to make sure the Promotheus client is actually sending these metrics before merging this PR

I believe the metrics should be sent here: https://earthangel-b40313e5.influxcloud.net/d/YS2KrS_nk/jbi?orgId=1

Are you looking to make charts and/or notification events from these metrics?

leplatrem:
Existing metrics are reported here:  https://earthangel-b40313e5.influxcloud.net/d/YS2KrS_nk/jbi?orgId=1
Let's see if they show up once this gets deployed

PR #139, opened 2022-07-22
Do not serialize raw Action objects and their callable (fixes #136, #137)

### LINKED ISSUE #136: Powered By API -  Returning 500
Screen Shot 2022-07-21 at 2 52 10 PM

### LINKED ISSUE #137: Whiteboard Tags API - 500 Internal Server Error
Screen Shot 2022-07-21 at 2 53 23 PM

leplatrem:
Fixes #137
Fixes #136 

I could reproduce these issues by:

- setting up real clients (Bugzilla API key, JIRA API key) in `.env`
- running locally with `make start`
- checking that clients can connect with `http :8000/__heartbeat__`

In order to force initialization of actions on the `/whiteboard_tags/` endpoint I added this:

```diff
 app.include_router(monitor_router)
@@ -109,6 +110,9 @@ def get_whiteboard_tags(
     actions: Actions = Depends(configuration.get_actions),
 ):
     """API for viewing whiteboard_tags and associated data"""
+    for action in actions:
+        assert callable(action.callable)  # will load Python module
+
```

And then, reaching out `/whiteboard_tags/` would crash, trying to serialize:
1. the action executor class (`callable` attribute)
2. the bugzilla client
3. some internal requests/pool/socket object

I could only fix it by using `.dict(exclude={"callable"})` (or using model config and just `.dict()` as in the present patch)

I failed at reproducing the situation in a unit test (naive version below):
```diff

+def test_whiteboard_tags_with_initialized_actions(
+    anon_client, mocked_jira, webhook_create_example
+):
+    assert configuration.get_actions(), "There is at least one configured action"
+    for action in configuration.get_actions():
+        action.callable(webhook_create_example)
+
+    resp = anon_client.get("/whiteboard_tags/")
+    assert resp.status_code == 200  # not crashing :)
+
```
because the issue comes from an internal urlib object that only gets initialized when session is established

### DIFF
--- a/src/app/api.py
+++ b/src/app/api.py
@@ -10,7 +10,6 @@ from typing import Dict, List, Optional
 import sentry_sdk
 import uvicorn  # type: ignore
 from fastapi import Body, Depends, FastAPI, Request
-from fastapi.encoders import jsonable_encoder
 from fastapi.responses import HTMLResponse
 from fastapi.staticfiles import StaticFiles
 from fastapi.templating import Jinja2Templates
@@ -112,7 +111,7 @@ def get_whiteboard_tags(
     if existing := actions.get(whiteboard_tag):
         filtered = {whiteboard_tag: existing}
     else:
-        filtered = actions.by_tag
+        filtered = actions.by_tag  # type: ignore
     return {k: v.dict() for k, v in filtered.items()}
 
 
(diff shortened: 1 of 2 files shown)


grahamalama (approved):

grahamalama:
Reproduced and verified this fix locally.

We should add an issue to reproduce this through a test (and I still think this can all be done through straight model config), but in the interest of getting this bug fixed, let's get this shipped.

PR #142, opened 2022-07-22
Use `exclude` keyword in dict method to exclude Action `callable` field

grahamalama:
As a follow-on to #139 (which addresses #136 and #137):

The model config approach we used in d46eb39 seemed to help with the `/powered_by_jbi` endpoint, but errors were still appearing in `/whiteboard_tags`.

We found that using the `action.dict(exclude=<...>)` method seemed to _actually_ exclude the `callable` field. ([pydantic docs](https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude))

It's unclear at this point why this method works and the model config approach doesn't.

### DIFF
--- a/src/app/api.py
+++ b/src/app/api.py
@@ -112,7 +112,7 @@ def get_whiteboard_tags(
         filtered = {whiteboard_tag: existing}
     else:
         filtered = actions.by_tag  # type: ignore
-    return {k: v.dict() for k, v in filtered.items()}
+    return {k: v.dict(exclude={"callable"}) for k, v in filtered.items()}
 
 
 @app.get("/jira_projects/")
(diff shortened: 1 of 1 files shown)


src/app/api.py
```diff
@@ -36,6 +36,7 @@
     title="Jira Bugzilla Integration (JBI)",
     description="JBI v2 Platform",
     version="3.0.2",
+    debug=True,
```
bsieber-mozilla, line 39: Do we want to make this an env var?

bsieber-mozilla (approved):
LGTM; minor comment added. 

Thanks for looking into this

src/app/api.py
grahamalama: 😱 we do, but that wasn't my intent for this PR. Good catch.

bsieber-mozilla: Feel free to resolve this and we can make the changes later!

PR #145, opened 2022-07-24
Move initialized action callable to private field

grahamalama:
When we included the callable in a public model field, we ran into a bug (#136 and #137) where `jsonable_encoder` couldn't serialize an action with an initialized `requests.Session` object.

Here's how I recreated the error we saw:

```diff
diff --git a/tests/unit/jbi/bugzilla_action.py b/tests/unit/jbi/bugzilla_action.py
new file mode 100644
index 0000000..d91e324
--- /dev/null
+++ b/tests/unit/jbi/bugzilla_action.py
@@ -0,0 +1,16 @@
+from bugzilla import Bugzilla
+from requests import Session
+
+session = Session()
+
+
+class TestBugzillaAction:
+    def __init__(self):
+        self.bz = Bugzilla(url=None, requests_session=session)
+
+    def __call__(self):
+        return lambda: "test"
+
+
+def init():
+    return TestBugzillaAction()
diff --git a/tests/unit/test_models.py b/tests/unit/test_models.py
new file mode 100644
index 0000000..6dc37ca
--- /dev/null
+++ b/tests/unit/test_models.py
@@ -0,0 +1,15 @@
+from src.jbi.models import Action
+from fastapi.encoders import jsonable_encoder
+
+
+def test_model_serializes():
+    action = Action.parse_obj(
+        {
+            "whiteboard_tag": "devtest",
+            "contact": "person@example.com",
+            "description": "test config",
+            "module": "tests.unit.jbi.bugzilla_action",
+        }
+    )
+    assert callable(action.callable)
+    jsonable_encoder(action)
```

For some reason, excluding the `callable` field with `Config` didn't seem to fix the bug. So instead, in this PR we move the callable to a private field with Pydantic's `PrivateAttr`. 

We also add a test case as a regression test for the bug we observed.

This seems like more of a bug with `jsonable_encoder`, but it's probably good to remove the callable from the dict representation of the model anyway.

### DIFF
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -8,7 +8,7 @@ from inspect import signature
 from types import ModuleType
 from typing import Any, Callable, Dict, List, Literal, Mapping, Optional, Set, Union
 
-from pydantic import EmailStr, Extra, Field, root_validator, validator
+from pydantic import EmailStr, Extra, Field, PrivateAttr, root_validator, validator
 from pydantic_yaml import YamlModel
 
 
@@ -24,13 +24,18 @@ class Action(YamlModel):
     enabled: bool = False
     allow_private: bool = False
     parameters: dict = {}
+    _caller: Callable = PrivateAttr(default=None)
 
-    @functools.cached_property
-    def callable(self) -> Callable:
-        """Return the initialized callable for this action."""
+    def _initialize_caller(self):
         action_module: ModuleType = importlib.import_module(self.module)
-        initialized: Callable = action_module.init(**self.parameters)  # type: ignore
-        return initialized
+        initialized_caller: Callable = action_module.init(**self.parameters)  # type: ignore
+        self._caller = initialized_caller
+
+    def call(self, payload):
+        """Run action callable"""
+        if not self._caller:
+            self._initialize_caller()
+        return self._caller(payload)
 
     @root_validator
     def validate_action_config(cls, values):  # pylint: disable=no-self-argument
@@ -56,7 +61,6 @@ class Action(YamlModel):
 
         extra = Extra.allow
         keep_untouched = (functools.cached_property,)
-        fields = {"callable": {"exclude": True}}
 
 
 class Actions(YamlModel):
--- a/src/jbi/models.py
+++ b/src/jbi/models.py
@@ -8,7 +8,7 @@ from inspect import signature
 from types import ModuleType
 from typing import Any, Callable, Dict, List, Literal, Mapping, Optional, Set, Union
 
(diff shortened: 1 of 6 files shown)


tests/unit/jbi/test_models.py
```diff
@@ -0,0 +1,35 @@
...
+    """Regression test to assert that action with initialzed Bugzilla client serializes"""
+    action = Action.parse_obj(
+        {
+            "whiteboard_tag": "devtest",
+            "contact": "person@example.com",
+            "description": "test config",
+            "module": "tests.unit.jbi.bugzilla_action",
+        }
+    )
+    action._initialize_caller()
```
leplatrem, line 18: Why call this private method, and not `call()`?

tests/unit/jbi/bugzilla_action.py
```diff
@@ -0,0 +1,16 @@
+from bugzilla import Bugzilla
+from requests import Session
+
+session = Session()
+
+
+class TestBugzillaAction:
```
leplatrem, line 7: Could we pick a name that doesn't have `Test` in it? (eg. `FakeBugzillaAction`)

src/jbi/models.py
```diff
@@ -24,13 +24,18 @@ class Action(YamlModel):
...
-    def callable(self) -> Callable:
-        """Return the initialized callable for this action."""
+    def _initialize_caller(self):
         action_module: ModuleType = importlib.import_module(self.module)
-        initialized: Callable = action_module.init(**self.parameters)  # type: ignore
-        return initialized
+        initialized_caller: Callable = action_module.init(**self.parameters)  # type: ignore
+        self._caller = initialized_caller
+
+    def call(self, payload):
```
leplatrem, line 34: Can't this be a property instead?

It confused my monday morning brain. 
We have two notions of actions. This one here, representing configuration entries, which would now have a `call()` method, and the whiteboard actions, which are callables, and would be called from here via the runner.

I think that if we could, it would be slightly easier to follow if the action model would only have attributes, and the runner be in charge to execute them.
If we can't, let's polish naming a bit maybe to see if it helps (we have `call()`, `__call__()`, `execute_action()`, ...)

grahamalama: Good thinking @leplatrem. Addressed this in d78c61f.

Also, luckily, turns out a `@property` isn't normally included in the serialized object. So it was really just a matter of "caching" the initialized caller in the private field (instead of using `keep_untouched` and `functools.cached_property`) to have the callable be a property.

Also, I renamed `callable` -> `caller` just because I didn't want to mirror Python's built-in `callable` function.

leplatrem (approved):

PR #151, opened 2022-07-25
Update config.nonprod.yaml

bsieber-mozilla:
A request was made for a `JB` JBI Bin for both mozit-test and mozilla-hub jira environments: https://mozilla-hub.atlassian.net/browse/SDD-7493

The intent is to create a project space for defaulting/testing.

By making the project key consistent in non-prod and prod--and having the jira environments aligned--it makes me wonder if we need multiple configuration files.

### DIFF
--- a/config/config.nonprod.yaml
+++ b/config/config.nonprod.yaml
@@ -2,10 +2,10 @@
 # Action Config
 - whiteboard_tag: nonprodtest
   contact: bsieber@mozilla.com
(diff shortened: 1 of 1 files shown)


leplatrem:
> it makes me wonder if we need multiple configuration files.

Let's keep them for now. If it becomes a hassle to duplicate values, we can decide later

leplatrem (approved):

PR #152, opened 2022-07-26
 Retry Jira and Bugzilla on error (fixes #33) 

### LINKED ISSUE #33: Enhance the "retry" capabilities of JBI when there are external service errors

leplatrem:
Fixes #33


bsieber-mozilla (approved):
LGTM

PR #155, opened 2022-07-27
Improve logging and status reporting of actions

leplatrem:
Without these changes, not all log entries would contain hints about the operation being executed.

In the GCP Log Explorer, when filtering on the `jsonPayload.Fields.operation` field, we can now see most of the log entries emitted by the actions. It will become a lot easier to follow what is going on for a certain incoming webhook.

I also replaced the `{"status": "..."}`  by a proper enum.

The runner logs with `INFO` level the main operation value returned by the action, instead of just `success`. When troubleshooting, we are likely to filter log entries with `operation != ignore`  rather than `operation = success`


grahamalama:
Questions about some implications of this PR (and I guess all the logging we've started to put into actions):
- Are action authors responsible for incorporating logging into their custom actions? Should include a directive for this in `src/jbi/whiteboard_actions/README.md`?
- Where before, action authors were only responsible for returning a dict of a response payload, now they're required to return an `Operation` and a payload. Can we document this in the `README` as well? Is there a way to enforce this say with mypy?
- `Tuple[Operation, Dict]` seems reused a few times, and I think would be the expected return type for all callers. Does it deserve its own type alias?

grahamalama:
Also, something I was thinking about when looking over this PR.

Instead of having authors be responsible for logging their own interactions with Jira and Bugzilla, what if we instead moved this logging "closer" to the services by making adapters?

"Back of the napkin" sketch, but I was thinking something like:

```python
from src.jbi import Operations


class JiraClientAdapter:
    def __init__(self, jira_client=get_jira()) -> None:
        self.jira_client = jira_client

    def create_issue(self, fields, payload):
        bug_object = payload.bugzilla_object
        log_context = {
            "request": payload.dict(),
            "bug": bug_object.dict(),
        }
        logger.debug(
            "Create new Jira issue for Bug %s",
            bug_object.id,
            extra={
                **log_context,
                "operation": Operations.CREATE,
            },
        )
        return self.jira_client.create_issue(fields=fields)
    ...
```

If not _eliminating_ the need to log from custom actions, it would at least put less of the burden on custom action authors.

leplatrem:
> Are action authors responsible for incorporating logging into their custom actions? Should we include a directive for this in src/jbi/whiteboard_actions/README.md?

I think they would be responsible yes. It's our responsilbility to make it convenient.

>  Can we document this in the README as well?

Will do.

> Does it deserve its own type alias?

Yeah, something like `ActionResult`

> what if we instead moved this logging "closer" to the services by making adapters?

I see what you're trying to achieve, but I would rather wrap the logger object than the client. For example, to add the notion of  "stacked" context, so that when a `bug` object is added the context, let's say by the runner, all subsequent logging calls in the action will have it too. 

I know they are libraries like `stuctlog` who provide abstractions around that idea. Could be worth investigating

bsieber-mozilla:
> I know they are libraries like stuctlog who provide abstractions around that idea. Could be worth investigating

As a reference, CTMS uses this library [here](https://github.com/mozilla-it/ctms-api/blob/cde9694e2992c2c9d8315ba9c6954c50a6facf37/ctms/log.py#L14).

src/jbi/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+    HANDLE = "handle"
+    EXECUTE = "execute"
+    IGNORE = "ignore"
+
+    CREATE = "create"
+    UPDATE = "update"
+    DELETE = "delete"
+    COMMENT = "comment"
+    LINK = "link"
```
bsieber-mozilla, line 21: Is it expected/implied that a stakeholder would be limited to these Operations?  Or are these open to alternate results? (Is this too constraining or the right amount of limited choice?)

I like the use of these Enum's instead of one-off strs; in particular the logging of the operations really seems to make sense.

src/jbi/whiteboard_actions/README.md
```diff
@@ -7,13 +7,21 @@ Let's create a `new_action`!
...
+        def execute(payload) -> ActionResult:
+            print(f"{optional_param}, going to {jira_project_key}!")
+            return Operation.CREATE, {"result": 42}
+
+        return execute
     ```
     1. In the above example the `jira_project_key` parameter is required
     1. `optional_param`, which has a default value, is not required to run this action
     1. `init()` returns a `__call__`able object that the system calls with the Bugzilla request payload
+    1. The returned operation will be used for logging and the along side value will be returned by the WebHook endpoint
```
bsieber-mozilla, line 23:
```suggestion
    1. The returned ActionResult features an `operation` (used for logging) along with a Dict (used as a response to the WebHook endpoint).
```

Maybe? Up to you though

src/jbi/runner.py
```diff
@@ -77,24 +69,24 @@ def execute_action(
             action.whiteboard_tag,
             action.module,
             bug_obj.id,
-            extra={"operation": Operations.EXECUTE, **log_context},
+            extra={"operation": Operation.EXECUTE, **log_context},
         )
 
-        content = action.caller(payload=request)
+        operation, details = action.caller(payload=request)
```
bsieber-mozilla, line 75: Thinking on the naming here... maybe "operation"  should have some implication that it's use is for logs?

src/jbi/__init__.py
bsieber-mozilla: I'm a bit fearful that "Operation" could be confusing/limiting for stakeholders--that is if there gets to be complex actions.

bsieber-mozilla (commented):
"Status" as an arbitrary str is perhaps too open a policy; however "Operation" seems to be a bit more limiting but appropriate.

Bugzilla's webhook endpoint is just looking for a 200--whatever else is supplied is ignored, iirc--and the results of action are primarily Pass/Fail (where Fail are exceptions)...so just throwing it out there:
```python
ActionResult = Tuple[Boolean, Dict]
```
or
```python
class ActionResult:
    success: Boolean
    log_context: Dict
```
- where `Dict` is a dict that will be appended to the log_context
- where `boolean` is True for Passing actions (making me question when will False be supplied? It could be explicitly provided--but maybe we should limit non-passing actions to err out as exceptions? If we only expect Pass=True, then the ActionResult might seemingly be just a Dict.)


Overall, I think this PR is an improvement for the existing system; if you wanted to merge this and then iterate on the design, I'm in support.

leplatrem:
I understand what you mean. Not that we won't be able to add new operations to the `Enum`, but that summarizing the whole action execution in a single operation is not always going to make sense (eg. if the action does `comment` + `link` + `update`...).

I like your proposal:
- `return True` if something happened
- `return False` if the action was a noop
- `raise` if error

bsieber-mozilla (approved):
LGTM; thanks for the changes!

PR #156, opened 2022-07-27
Start (naive) troubleshooting section in README

leplatrem:


README.md
```diff
@@ -219,6 +219,59 @@ GET /whiteboard_tags/
 }
 ```
 
+## Troubleshooting
+
+### Bugzilla tickets are now showing up as issues on Jira
```
grahamalama, line 224: are _not_ showing up?

README.md
```diff
@@ -219,6 +219,59 @@ GET /whiteboard_tags/
...
 
+## Troubleshooting
+
+### Bugzilla tickets are not showing up as issues on Jira
+
+Checklist:
+
+- The Bugzilla Webhook is setup to be executed for your project
+- The Bugzilla Webhook is still **enabled**
+- The whiteboard tag is listed in `GET /whiteboard_tags/`
```
bsieber-mozilla, line 230:
```suggestion
- The whiteboard tag is enabled in `GET /whiteboard_tags/`
```
Maybe enabled, as listed ones might be disabled?

bsieber-mozilla (approved):
Thanks for starting this doc!

README.md
```diff
@@ -219,6 +219,59 @@ GET /whiteboard_tags/
...
+- The Jira project is listed in `GET /jira_projects/`
```
mostlygeek, line 231: Would an explicit action be easier: `open https://jbi.services.mozilla.com/powered_by_jbi/` and check if your project is listed there.

README.md
```diff
@@ -219,6 +219,59 @@ GET /whiteboard_tags/
...
+
+For private bugs:
+
+- The Bugzilla account has permissions to read the private bugs
+
+### Log Explorer Queries Examples
```
mostlygeek, line 237: These queries are great time savers! 
I think having them part of troubleshooting scenarios would make them easier to understand.  For example: 

## Scenario:  My bug was updated but the data is not syncing? 

Step 1: Check if Bugzilla sent anything to JBI for your bug: 

```
jsonPayload.Fields.bug.id=1780798
```

Step 2: If nothing found, check Bugzilla web hook settings.  The cause may be the bugzilla project was not configured to call the webhook. 

## Scenario: JBI was unsuccessful updating Jira 
...

## Scenario:  ...

leplatrem: Yes, I agree! 

But from the latest conversations with Katie, it doesn't seem like consumers are likely to have access to Log queries. I will leave them as example, and later we can rewrite them in a more narrative way if it becomes relevant

leplatrem:
Thanks for your feedback 🙏

PR #159, opened 2022-07-27
Show 'Bug XXXX' instead of Bugzilla Ticket in Jira link

leplatrem:
Screenshot 2022-07-27 at 13 56 17

In the Firefox world, bugs are always referred as `Bug <number>`

### DIFF
--- a/src/jbi/whiteboard_actions/default.py
+++ b/src/jbi/whiteboard_actions/default.py
@@ -215,7 +215,7 @@ class DefaultExecutor:
         jira_response = self.jira_client.create_or_update_issue_remote_links(
             issue_key=jira_key_in_response,
             link_url=bugzilla_url,
-            title=f"Bug {bug_obj.id}",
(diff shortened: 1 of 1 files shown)


src/jbi/whiteboard_actions/default.py
```diff
@@ -215,7 +215,7 @@ def create_and_link_issue(
         jira_response = self.jira_client.create_or_update_issue_remote_links(
             issue_key=jira_key_in_response,
             link_url=bugzilla_url,
-            title="Bugzilla Ticket",
```
grahamalama, line 218: Would it be redundant to include the context that it's a Bugzilla bug? `Bugzilla Bug XXXXXXX`

grahamalama (approved):

src/jbi/whiteboard_actions/default.py
leplatrem: Yeah, you're right

PR #160, opened 2022-07-27
Use statsd client instead of Promotheus

leplatrem:


bin/docker-entrypoint.sh
```diff
@@ -2,15 +2,5 @@
 
 set -e
 
-# setup an empty directory for Prometheus metrics
-if [ ${PROMETHEUS_MULTIPROC:-0} -eq 1 ]; then
-    prometheus_multiproc_dir=`mktemp -td prometheus.XXXXXXXXXX` || exit 1
-    export prometheus_multiproc_dir
-fi
```
grahamalama, line 9: Do we even need this file anymore?

grahamalama (approved):
Do we need `bin/docker-entrypoint.sh](https://github.com/mozilla/jira-bugzilla-integration/pull/160/files#diff-368e25a43c4a3200ecb9ae21e7b2a94da05d8eb24680bd6c2a37ee0e275bcd0a)

leplatrem:
> where will these status go?

They will go to our Grafana instance. Once they show up, we can add panels to our nonprod/prod dashboards

PR #161, opened 2022-07-27
Tweak logging (log `500` requests), other small logging-adjacent tweaks

grahamalama:
Commits and the final product of this PR are definitely in a WIP state, but I'd like feedback now to make sure we're on the right path for our logging config.

### DIFF
--- a/asgi.py
+++ b/asgi.py
@@ -1,0 +1,15 @@
+import uvicorn
+
+from src.app.environment import get_settings
+
+settings = get_settings()
+
+
+if __name__ == "__main__":
+    uvicorn.run(
+        "src.app.api:app",
+        host=settings.host,
+        port=settings.port,
+        reload=settings.app_reload,
+        log_level=settings.log_level,
+    )
--- a/asgi.py
+++ b/asgi.py
@@ -1,15 +1,19 @@
 import uvicorn
 
 from src.app.environment import get_settings
+from src.app.log import CONFIG as LOGGING_CONFIG
 
 settings = get_settings()
 
 
 if __name__ == "__main__":
-    uvicorn.run(
-        "src.app.api:app",
-        host=settings.host,
-        port=settings.port,
-        reload=settings.app_reload,
-        log_level=settings.log_level,
+    server = uvicorn.Server(
+        uvicorn.Config(
+            "src.app.api:app",
+            host=settings.host,
+            port=settings.port,
+            reload=settings.app_reload,
+            log_config=LOGGING_CONFIG,
+        )
     )
+    server.run()
--- a/src/app/environment.py
+++ b/src/app/environment.py
@@ -2,11 +2,20 @@
 Module dedicated to interacting with the environment (variables, version.json)
 """
 import json
+from enum import Enum
 from functools import lru_cache
 from pathlib import Path
 from typing import Optional
 
 from pydantic import AnyUrl, BaseSettings
+
+
+class Environment(str, Enum):
+    """Production environment choices"""
+
+    LOCAL = "local"
+    NONPROD = "nonprod"
+    PRODUCTION = "production"
 
 
 class SentryDsn(AnyUrl):
@@ -19,8 +28,8 @@ class Settings(BaseSettings):
     host: str = "0.0.0.0"
     port: int = 8000
     app_reload: bool = True
-    env: str = "nonprod"
     max_retries: int = 3
+    env: Environment = Environment.NONPROD
 
     # Jira
     jira_base_url: str = "https://mozit-test.atlassian.net/"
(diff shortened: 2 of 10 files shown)


src/app/log.py
```diff
@@ -26,13 +27,43 @@ def configure_logging():
...
+            },
         },
         "loggers": {
-            "request.summary": {"handlers": ["console"], "level": "INFO"},
-            "backoff": {"handlers": ["console"], "level": "INFO"},
-            "src.jbi": {"handlers": ["console"], "level": "DEBUG"},
+            "request.summary": {
+                "handlers": ["console"],
+                "level": logging.INFO,
+                "propagate": False,
```
bsieber-mozilla, line 39: What will `"propagate": False,`  do here?

bsieber-mozilla (commented):
Thanks for this refactoring, I like the asgi.py addition!

src/app/log.py
leplatrem: It will prevent the log messages emitted with as `request.summary` to be caught by loggers called `"request"` or `""` (=`root`)

src/app/log.py
```diff
@@ -14,6 +14,7 @@ def configure_logging():
     """Configure logging; mozlog format for Dockerflow"""
     logging_config = {
         "version": 1,
+        "disable_existing_loggers": False,
```
leplatrem, line 17: We want to disable existing loggers, otherwise we don't control what loggers are configured out there

src/app/log.py
```diff
@@ -26,13 +27,43 @@ def configure_logging():
...
-            "backoff": {"handlers": ["console"], "level": "INFO"},
-            "src.jbi": {"handlers": ["console"], "level": "DEBUG"},
+            "request.summary": {
+                "handlers": ["console"],
+                "level": logging.INFO,
+                "propagate": False,
+            },
+            "backoff": {
+                "handlers": ["console"],
+                "level": logging.INFO,
```
leplatrem, line 43: The default logging level of `backoff` is already `INFO`, I know I'm the one that put it here, but we can remove it actually

src/app/log.py
```diff
@@ -26,13 +27,43 @@ def configure_logging():
...
+                "propagate": False,
+            },
+            "src": {
+                "handlers": ["console"],
+                "level": settings.log_level.upper(),
```
leplatrem, line 48: The `level` param (and on the following below) is not necessary, since they all use the `console` handler, which is already set to `settings.log_level.upper()`

src/app/environment.py
```diff
@@ -2,13 +2,20 @@
...
 import json
+from enum import Enum
 from functools import lru_cache
 from pathlib import Path
 from typing import Optional
 
 from pydantic import AnyUrl, BaseSettings
 
 
+class Environment(str, Enum):
```
leplatrem, line 13: 👀 I like this more than https://github.com/mozilla/jira-bugzilla-integration/blob/a888e9fa080791e29b97e3284980108875c5186c/src/jbi/__init__.py#L8

asgi.py
```diff
@@ -0,0 +1,19 @@
...
+
+from src.app.environment import get_settings
+from src.app.log import CONFIG as LOGGING_CONFIG
+
+settings = get_settings()
+
+
+if __name__ == "__main__":
+    server = uvicorn.Server(
+        uvicorn.Config(
```
grahamalama, line 11: I used this `Server(Config())` approach because I found that when using just `uvicorn.run()`, some exception was raised when trying to set `reload` and `log_config` simultaneously.

asgi.py
```diff
@@ -0,0 +1,19 @@
...
+            "src.app.api:app",
+            host=settings.host,
+            port=settings.port,
+            reload=settings.app_reload,
+            log_config=LOGGING_CONFIG,
```
grahamalama, line 16: https://www.uvicorn.org/settings/#logging

I think this has the same effect as `logging.config.dictConfig`

src/app/log.py
grahamalama: I thought the general advice here was to set this to `False` and manually / explicitly handle any loggers?

leplatrem: Saw your notes, yes you're right

grahamalama: For posterity, these were some sources that guided me to set this to `False`:
- https://guicommits.com/how-to-log-in-python-like-a-pro/
> I recommend keeping disable_existing_loggers False to avoid your system from swallowing other unexpected issues that might happen. If you want to modify some other loggers I recommend (although boring) overwriting them explicitly!
- https://docs.djangoproject.com/en/4.0/topics/logging/#configuring-logging
> Thus you should be very careful using 'disable_existing_loggers': True; it’s probably not what you want. Instead, you can set disable_existing_loggers to False and redefine some or all of the default loggers

PR #162, opened 2022-07-28
Run app as a single uvicorn process (fixes #133)

### LINKED ISSUE #133: Run app as a single process in production
From the FastAPI [docs](https://fastapi.tiangolo.com/deployment/docker/?h=deployment#replication-number-of-processes):

> If you have a cluster of machines with Kubernetes (...) then you will probably want to handle replication at the cluster level instead of using a process manager (like Gunicorn with workers) in each container.
>...
> In those cases, you would probably want to (...) [run a] single Uvicorn process instead of running something like Gunicorn with Uvicorn workers.

I think this means we should:
- change the default container CMD to run `uvicorn`
- remove `gunicorn_conf.py`

leplatrem:
Fixes #133

This could wait the `asgi.py`  module from #161


bsieber-mozilla (approved):
LGTM! 

👋 g🦄

bsieber-mozilla:
> Fixes #133
> 
> This could wait the `asgi.py` module from #161

I think since the referenced PR has been merged there might need to be a change to this PR.

Dockerfile
```diff
@@ -41,4 +41,4 @@ WORKDIR /app
 COPY . .
 
 EXPOSE $PORT
-CMD ["gunicorn", "-k", "uvicorn.workers.UvicornWorker", "-c", "config/gunicorn_conf.py", "src.app.api:app"]
+CMD ["python", "-m", "asgi"]
```
bsieber-mozilla, line 44: @leplatrem, this was the minor commit: "asgi"

Issue #140, opened 2022-07-22
`get_jira()`  and `get_bugzilla()`  should be cached


leplatrem:
Otherwise, a session is established on every instantiation. For example `GET __heartbeat__` takes more than 5 seconds on each call!

For actions, since the callable (ActionExecutor instance) is cached, it is already indirectly cached

PR #163, opened 2022-07-28
Retrieve comments from the API if they are private. Fixes #154

### LINKED ISSUE #154: Creating a patch attachment in bugzilla posts an empty comment to Jira
Here I posted a patch to phabricator for a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1780798#c1
JBI commented in the Jira ticket with "None": https://mozilla-hub.atlassian.net/browse/MR2-3072?focusedCommentId=573910

Mossop:


src/jbi/whiteboard_actions/default.py
```diff
@@ -69,15 +69,17 @@ def comment_create_or_noop(self, payload: BugzillaWebhookRequest):
...
-            issue_key=linked_issue_key,
-            comment=payload.map_as_jira_comment(),
-        )
-        logger.debug(
-            "Comment added to Jira issue %s",
-            linked_issue_key,
-            extra=log_context,
-        )
+        comment = payload.map_as_jira_comment()
+        if comment is not None:
```
leplatrem, line 73: I would invert the condition and early return `{"status": "noop"}`

src/jbi/whiteboard_actions/default.py
```diff
@@ -69,15 +69,17 @@ def comment_create_or_noop(self, payload: BugzillaWebhookRequest):
...
+            jira_response = self.jira_client.issue_add_comment(
+                issue_key=linked_issue_key,
+                comment=payload.map_as_jira_comment(),
+            )
+            logger.debug(
+                "Comment added to Jira issue %s",
+                linked_issue_key,
+                extra=log_context,
+            )
         return {"status": "comment", "jira_response": jira_response}
```
leplatrem, line 83: If this doesn't crash about `jira_response`, this line is probably not executed

src/jbi/bugzilla.py
```diff
@@ -227,6 +227,15 @@ def map_as_jira_comment(self):
         comment: BugzillaWebhookComment = self.bug.comment
         commenter: BugzillaWebhookUser = self.event.user
         comment_body: str = comment.body
+
+        if comment.is_private:
+            bug_comments = get_bugzilla().get_comments([self.bug.id])
```
leplatrem, line 232: I don't think we should call `get_bugzilla()`  from here. It does not feel like it's the right way.

I would try something like that:
```py
if comment.is_private:
     all_comments = self.bugzilla_object.fetch_all_comments()
     matching_comments = [c for c in all_comments if c["id"] == comment.id]
     ...
```


I now realize we need to rename these  *map* methods, because the current state is confusing, we have `map_as_jira_comment` and `map_as_comment`. And while reading your PR it took me some time to understand them. One seems to be `bugzilla_comment_to_jira_comment`  and the other one `bugzilla_change_as_jira_comment` (not saying these are great names).

src/jbi/whiteboard_actions/default.py
Mossop: I theory it should never happen that we can't get the comment from bugzilla (unless maybe the comment can be deleted between the webhook and the attempt to retrieve?). Maybe it would be better to raise an exception in that case?

src/jbi/bugzilla.py
leplatrem: > so I'm not sure what the improvement is there

In terms of readability, it would be a little improvement, but in terms of performance, you are totally right actually. The reason I was nervous about this is #140 and we can tackle that in another PR.

src/jbi/whiteboard_actions/default.py
leplatrem: What I meant is that if comment is None then `jira_response` is not defined

Mossop:
Ok I think this is ready for review again then?

tests/unit/jbi/whiteboard_actions/test_default.py
```diff
@@ -112,6 +112,21 @@ def test_added_comment(webhook_comment_example: BugzillaWebhookRequest, mocked_j
...
+    callable_object = default.init(jira_project_key="")
+
+    value = callable_object(payload=webhook_private_comment_example)
+
+    mocked_jira().issue_add_comment.assert_called_once_with(
+        issue_key="JBI-234",
+        comment="*(mathieu@mozilla.org)* commented: \n{quote}hello{quote}",
+    )
+    assert value["status"] == "comment"
+
```
leplatrem, line 128: Would you mind adding a tiny test where the private comment is not found (to test the `noop`  code path)?

```py
mocked_bugzilla().get_comments.return_value = {
     "bugs": {
         str(webhook_private_comment_example.bug.id): {
            "comments": []
         }
      }
 }     
```

leplatrem (approved):
LGTM! 

Having #140 would help here because calls to `get_bugzilla()` take time (authentication)

tests/unit/jbi/whiteboard_actions/test_default.py
Mossop: Done.

leplatrem:
🎉

PR #172, opened 2022-08-01
Create standardized release template

grahamalama:
Create a [release notes template](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes) so that we have standardized sections for each release. 

This method kind of encourages us to make better use of labels as well. I'm guessing that a section heading defined in this template will only show up if there's a PR that has that label.  Though it's easy enough to copy from the "Other Changes" section and paste to the correct section heading.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,26 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+  categories:
+    - title: Breaking Changes
+      labels:
+        - "breaking-change"
+    - title: Bug Fixes
+      labels:
+        - "bug"
+    - title: New Features
+      labels:
+        - "enhancement"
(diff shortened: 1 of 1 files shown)


.github/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - title: Configuration
+      labels:
+        - "configuration"
+    - title: Documentation
+      labels:
+        - "documentation"
+    - title: Dependency Updates
+      labels:
+        - "dependencies"
+    - title: Other Changes
```
leplatrem, line 24: I sometimes call this "Internal Changes", like, not affecting consumers/operators

leplatrem (approved):

PR #174, opened 2022-08-01
Break up `src/app/api.py` module

grahamalama:
The goal of this PR of to move some of the code currently present in `src/app/api.py` into more appropriate modules:

- moves the `request.summary` log field helper into `src/app/log.py`
- moves all route handlers into a new `src/app/router.py`
- renames `src/app/api.py` -> `src/app/main.py`
  I did want to name this module `app.py` since we're building a FastAPI `app`, but since it's in an `app/` package, I went with the `main` name.

The result of this PR is that the three modules involved have more obvious "jobs:
- `log.py`: logging config and any helpers
- `router.py`: all route handling logic
- `main.py`: building the FastAPI app (middleware, adding static files, registering the router)


leplatrem (approved):
Well done, I like it better too.

The tests filenames still match? `test_api.py` covers `app/main.py` and `app/router.py`. I think it's ok.

grahamalama:
> The tests filenames still match? test_api.py covers app/main.py and app/router.py. I think it's ok.

Nah that's a good call -- I like to have the test file names and src file names line up

leplatrem (approved):
I let you merge ;)

PR #176, opened 2022-08-02
Move most modules to package at root level, rename from "src" to "jbi"

grahamalama:
Before, we had this project structure
```
src
├── __init__.py
├── app
├── jbi
├── static
├── templates
```

the differences between `src/app` and `src/jbi` weren't clear enough
to warrant separate packages, so in this PR we move most modules
into the root package, renamed to `jbi/`. We also change the structure
of the test files accordingly and update any other files that referred
to src/* files.


jbi/log.py
```diff
@@ -37,7 +37,7 @@
     "loggers": {
         "": {"handlers": ["console"]},
         "request.summary": {"level": logging.INFO},
-        "src": {"level": logging.DEBUG},
```
leplatrem, line 40: 💫

leplatrem (approved):

PR #183, opened 2022-08-02
Merge `bugzilla.py` module with `models.py`

grahamalama:
Since `bugzilla.py` was a collection of Pydantic `BaseModel`s, in this PR we move them into `models.py` so that it is the one source of all models used in the app.


jbi/services.py
```diff
@@ -1,15 +1,19 @@
 """Services and functions that can be used to create custom actions"""
+from __future__ import annotations
```
bsieber-mozilla, line 2: What is`annotations` used for?

jbi/services.py
```diff
@@ -1,15 +1,19 @@
...
 import backoff
 import bugzilla as rh_bugzilla
 from atlassian import Jira, errors
 from statsd.defaults.env import statsd
 
 from jbi import environment
-from jbi.models import Actions
+
+if TYPE_CHECKING:
+    from jbi.models import Actions
```
bsieber-mozilla, line 16: Only import on `TYPE_CHECKING`? (What value is TYPE_CHECKING? are we expecting this as False?)

bsieber-mozilla (approved):
Minor resolvable comments, overall lgtm!

grahamalama:
@bsieber-mozilla 
> What is annotations used for?

> Only import on TYPE_CHECKING? (What value is TYPE_CHECKING? are we expecting this as False?)

I was running into a circular import after making this change, and the only reason for that was for the purposes of type hinting. I found [this](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/) blog post that pointed me to add the changes that you called out in your review.

Issue #191, opened 2022-08-03
How to determine Total Count vs Success Count of Metrics/Instrumented Client?


bsieber-mozilla:
https://github.com/mozilla/jira-bugzilla-integration/blob/fcb739081a5c56ad39a4d4cf2bc962a1c0ec295a/jbi/services.py#L47-L52

The '.count' element here appears to be a total count

bsieber-mozilla:
In #173 we'd like to be able to understand and display the total count of method split based on success/failure of the call.

leplatrem:
From what I see in https://github.com/mozilla/jira-bugzilla-integration/issues/173#issuecomment-1204331174 we are going to use logs instead of StatsD metrics.
So we should probably just add a log message around these too

bsieber-mozilla:
Are we looking to log the jira and bugzilla api calls along with other details or just the calls?

bsieber-mozilla:
General question: 
Do we have enough as is to collect the desired data? Are we looking to add more logs/metrics?

bsieber-mozilla:
https://mozilla-hub.atlassian.net/browse/SVCSE-707

Issue #173, opened 2022-08-01
Provide metrics to measure the amount and measure the type of work done by JBI


mostlygeek:
I'm interesting in being able to measure how much work manual JBI is automating away for us. 

I believe the numbers I need for this are: 

* number of bugzilla operations done, segmented by type and status (success, failed, etc)
* number of jira operations done, segmented by type and status (success, failed, etc)
* I would also like to sum up those values into a monthly count 

Questions: 

* how much effort would it take to make that happen? 
* any recommendations on different metrics?

leplatrem:
Since v3.1.0 we have these metrics:

- `jbi.bugzilla.ignored.count`
- `jbi.bugzilla.processed.count`
- `jbi.action.execution.timer`

- `jbi.jira.methods.update_issue_field.count`
- `jbi.jira.methods.set_issue_status.count`
- `jbi.jira.methods.issue_add_comment.count`
- `jbi.jira.methods.create_issue.count`

- `jbi.jira.methods.update_issue_field.timer`
- `jbi.jira.methods.set_issue_status.timer`
- `jbi.jira.methods.issue_add_comment.timer`
- `jbi.jira.methods.create_issue.timer`

- `jbi.bugzilla.methods.getbug.count`
- `jbi.bugzilla.methods.get_comments.count`
- `jbi.bugzilla.methods.update_bugs.count`

- `jbi.bugzilla.methods.getbug.timer`
- `jbi.bugzilla.methods.get_comments.timer`
- `jbi.bugzilla.methods.update_bugs.timer`

Most of the measurements you describe should be doable with Grafana (we don't count failed calls, but since a Sentry is sent each time, we expect them to be 0).

In case Grafana is not adequate,  I think that the application logs are also sent to BigQuery. So in theory, with the appropriate permissions, we could also have a Redash sheet that queries logs to count operations (we log them consistently)

mostlygeek:
What I would like to do is be able to get a sum of the __change__ operations JBI has done over a specific time.  

In pseudo SQL: 

```
SELECT
  jbi_operation_name,
  year(timestamp), 
  month(timestamp),
  sum( if status="success", 1, 0 ) success,
  sum( if status="success", 1, 0 ) failed,
FROM jbi_logs
GROUP BY 
  jbi_operation_name,
  year(timestamp), 
  month(timeatmp)
```

To produce a table like: 

```
jbi.jira.set_issue_status    | 2022 | 8 | 100 | 0 |
[...]

PR #192, opened 2022-08-04
Adding additional metrics during instrumentation of clients (#191)

bsieber-mozilla:
In this PR we can now push metrics of successful jira calls and exceptions 

@leplatrem does this make sense? Can success rate already be determined by existing metrics?

### DIFF
--- a/jbi/services.py
+++ b/jbi/services.py
@@ -45,11 +45,20 @@ class InstrumentedClient:
             max_tries=settings.max_retries + 1,
         )
         def wrapped_func(*args, **kwargs):
+            metric_label = f"jbi.{self.prefix}.methods.{attr}"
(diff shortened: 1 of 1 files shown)


leplatrem:
> @leplatrem does this make sense? Can success rate already be determined by existing metrics?

I asked Ben [here](https://github.com/mozilla/jira-bugzilla-integration/issues/173#issuecomment-1205317252) whether having these high level operations are enough (instead of client SDK methods calls). 

If yes, then we should probably only add a `fail` operation, that we log when an exception is raised in runner.
If not, then maybe adjust this PR to add logger calls instead of StatsD metrics, since BigQuery seems to be the way to go for this kind of reporting.

PR #199, opened 2022-08-08
Avoid unnecessary typing imports

grahamalama:
- Use collection keyword instead of `typing.{Dict|List|Tuple|Set}` ([Since Python 3.9](https://peps.python.org/pep-0585/))
- replace `typing.Union` with `|` ([Since Python 3.10](https://peps.python.org/pep-0604/))


bsieber-mozilla (approved):
LGTM

PR #205, opened 2022-08-15
Refactor some test fixtures / setup

grahamalama:
This PR does several things that refactor our testing setup:
##### Decouple test objects from mocked service side effects
In some of our fixtures, we'd return an example test object, but also specify the returns of mocked side effects when that fixture was used. This made it slightly more challenging to modify test fixtures. It also wasn't obvious that the service mocking was happening when reading tests. 

Now, we set the return values of mocked services in tests. This makes tests a bit more verbose, but also makes them more explicit and obvious.

##### Add a `tests/fixtures/factories.py` module to facilitate easy test object creation

These factory functions aren't perfect, and eventually I'd like to move to using something like [pydantic-factories](https://github.com/Goldziher/pydantic-factories), but for now this is a good-enough approach to easily make different example objects in tests.

##### Move some action tests from `test_configuration.py` to `test_models.py`

Some of the Action tests in `test_configuration.py` were actually testing the Action validators, so it seemed more appropriate to have them live in `test_models.py`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,7 @@ from jbi.app import app
 from jbi.environment import Settings
 from jbi.models import Action, Actions, BugzillaWebhookComment, BugzillaWebhookRequest
 from tests.fixtures.factories import (
+    action_factory,
     bug_factory,
     webhook_event_factory,
     webhook_factory,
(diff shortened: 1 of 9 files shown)


bsieber-mozilla (approved):
> ##### Decouple test objects from mocked service side effects

Thanks for this decoupling.

> ##### Add a `tests/fixtures/factories.py` module to facilitate easy test object creation

I love the use of a factory pattern here, good call.

> ##### Move some action tests from `test_configuration.py` to `test_models.py`
👍 

---

LGTM! 🚢

PR #206, opened 2022-08-16
Allow disabling syncing whiteboard fields to Jira. Fixes #201

### LINKED ISSUE #201: JBI blows away any non-bugzilla labels set on issues in Jira
_This seems to be a concern to a stakeholder, capturing this as an issue to discuss and prioritize._

Screen Shot 2022-08-11 at 11 04 31 AM

Slack ref: https://mozilla.slack.com/archives/C016BTJKM9B/p1660239831970749

Mossop:
For projects that make use of the Jira labels field JBI currently overwrites those labels everytime the synced bug changes. This allows disabling the syncing of the whiteboard fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,9 @@ It will also set the Jira issue URL in the Bugzilla bug `see_also` field.
 - `jira_project_key`
     - string
     - The Jira project identifier
+- `sync_whiteboard_labels`
+    - boolean
+    - Whether to sync the Bugzilla status whiteboard labels to Jira. Defaults to `true`.
 
 Example configuration:
 ```yaml
@@ -83,6 +86,17 @@ Example configuration:
     module: jbi.actions.default
     parameters:
       jira_project_key: EXMPL
+```
+
+Example configuration that disables setting labels in Jira:
+```yaml
+    whiteboard_tag: example
+    contact: example@allizom.com
+    description: example configuration
+    module: jbi.actions.default
+    parameters:
+      jira_project_key: EXMPL
+      sync_whiteboard_labels: false
 ```
 
 ### Default with assignee and status action
@@ -101,6 +115,9 @@ If configured, the action supports setting the Jira issues's status when the Bug
 - `jira_project_key`
     - string
     - The Jira project identifier
+- `sync_whiteboard_labels`
+    - boolean
+    - Whether to sync the Bugzilla status whiteboard labels to Jira. Defaults to `true`.
 - `status_map` (optional)
     - mapping [str, str]
     - If defined, map the Bugzilla bug status to Jira issue status
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -1,10 +1,13 @@
 """
 Default action is listed below.
 `init` is required; and requires at minimum the `jira_project_key` parameter.
+The `label_field` parameter configures which Jira field is used to store the
+labels generated from the Bugzilla status whiteboard.
 
 `init` should return a __call__able
 """
 import logging
+from typing import Any
 
 from jbi import ActionResult, Operation
 from jbi.environment import get_settings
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -101,6 +116,10 @@ If configured, the action supports setting the Jira issues's status when the Bug
 - `jira_project_key`
     - string
     - The Jira project identifier
+- `label_field`
+    - string or null
+    - The Jira field to store the Bugzilla status whiteboard labels in, use `null` to disable.
+      Defaults to `labels`
```
bsieber-mozilla, line 122: What different possible Jira fields can the labels be stored in? (Do different Jira projects store labels in different issue elements?)

Mossop: Projects can add custom fields so theoretically any custom field that is added. It has turned out though that we've decided for our project to just not store the labels at all so I could change this setting to a simple boolean to turn it off if that is preferable.

bsieber-mozilla: I think an optional parameter that can be toggled would be my preference; @grahamalama / @leplatrem what are your thoughts here?

Mossop: I've made that change. If a different project ends up wanting them somewhere else we can address it then.

grahamalama (commented):
A few comments before I do a full review:
- can we add a description to the PR?
- In another comment above, you said:

> ... It has turned out though that we've decided for our project to just not store the labels at all so I could change this setting to a simple boolean to turn it off if that is preferable.

If no projects are using this setting, is it still worth merging this change? Are JBI users surprised by or not happy with the behavior as it currently is?

- If this is something we still want to change, have we considered an approach where we merge the labels? So essentially appending the Bugzilla whiteboard tags to the Jira issue rather than overwriting them?

Mossop:
> A few comments before I do a full review:
> 
>     * can we add a description to the PR?

Done

>     * In another comment above, you said:
> 
> 
> > ... It has turned out though that we've decided for our project to just not store the labels at all so I could change this setting to a simple boolean to turn it off if that is preferable.
>
> If no projects are using this setting, is it still worth merging this change? Are JBI users surprised by or not happy with the behavior as it currently is?

The previous version of the PR allowed both disabling syncing or syncing to a different field in Jira. I had thought we were going to do the latter but it turns out we just don't want the labels synced at all so we still want that functionality.

I should probably add it to the config for our project as part of this...

>     * If this is something we still want to change, have we considered an approach where we merge the labels? So essentially appending the Bugzilla whiteboard tags to the Jira issue rather than overwriting them?

Merging in the generic case has some challenges. You would need to retrieve the existing labels from Jira and attempt to determine which were set from Bugzilla previously. At least for private bugs the webhook request doesn't include the old value of the status whiteboard when it changes so that would involve querying the bug's history to figure that out. Then you can identify the Jira only labels and add them to the new Bug labels.

There are some ways to avoid that complexity, you could use some kind of per-project whitelist of labels to retain for instance. But either way you have a non-atomic update of the Jira fields that could still cause you to blow away fields set in between retrieving the old labels and setting the new.

Given that for our project we aren't interested in having the bug labels in Jira it seems a lot simpler to just be able to disable them.

jbi/actions/default.py
```diff
@@ -25,17 +28,20 @@
...
 
 
 class DefaultExecutor:
     """Callable class that encapsulates the default action."""
 
-    def __init__(self, jira_project_key, **kwargs):
+    def __init__(self, jira_project_key, sync_labels=True, **kwargs):
         """Initialize DefaultExecutor Object"""
         self.jira_project_key = jira_project_key
+        self.sync_labels = sync_labels
```
grahamalama, line 44: Another option could be to omit the `sync_labels` params in `init` and `__init__` and instead do
```suggestion
        self.sync_labels = kwargs.get("sync_labels", True)
```

not strictly asking for a change here, just a suggestion.

grahamalama (approved):

grahamalama (approved):

Issue #202, opened 2022-08-13
Change Jira's Bugzilla links to use the Bugzilla favicon and URL instead of a generic globe icon and "Bugzilla Ticket" title


cpeterso:
Enhancement suggestion:

When JBI links a Jira issue to its corresponding Bugzilla bug, the Jira issue's web link has a default globe icon and the generic title `Bugzilla Ticket`:

![image](https://user-images.githubusercontent.com/629193/184459591-0865de09-3987-48ae-8710-f54d14855408.png)

I think this makes visually finding the Bugzilla link on the busy Jira page harder than it needs to be. I think it would be more useful if the Bugzilla web link's favicon was Bugzilla's green bug favicon and showed the bug's full Bugzilla URL (or something descriptive like `Bugzilla bug 1234567`) instead of `Bugzilla Ticket`. The Bugzilla favicon and URL are shown by default when a user manually adds a Bugzilla link to a Jira issue:

![image](https://user-images.githubusercontent.com/629193/184459622-544ade8f-63dd-434e-aa59-a13c3feccf48.png)

mostlygeek:
@bsieber-mozilla any idea if this is even possible via the Jira API?

grahamalama:
Hmm, is this a regression? https://github.com/mozilla/jira-bugzilla-integration/pull/159

bsieber-mozilla:
@mostlygeek we've got some options to enhance the look of the remote-link.

I think it might be best to scrap adding a title to the remote-link and instead just use the URL? I think that jira might be capable enough to see a "bugzilla" domain and add the favicon (like the manual approach).

grahamalama:
@bsieber-mozilla it looks like that's what's still on `main` [at the moment](https://github.com/mozilla/jira-bugzilla-integration/blob/a9e3d2cdae3c5cf5ea740e5766a0397dcf77748a/jbi/actions/default.py#L270).

Maybe this ticket just hasn't been updated in a while, so that why the old `Bugzilla Ticket` link title is still there?

But I do like
> scrap adding a title to the remote-link and instead just use the URL

If that means the favicon will show up

bsieber-mozilla:
It looks like title and url are required. I'll do a local test to see if having the title=url will pull this thumbnail for users. Otherwise, there are more advanced properties we can modify with the API.

[...]

PR #207, opened 2022-08-16
Update to remote-link title; use URL, set icon favicon url.

bsieber-mozilla:
#202 

After testing locally, this doesn't add the favicon--but is one step closer to what was previously expected.

The atlassian-python-api does not accept "icon" as part of the api, however if we use the Jira REST API, we should be able to add a link to a favicon.

### DIFF
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -273,11 +273,13 @@ class DefaultExecutor:
             jira_key_in_response,
             extra=log_context.update(operation=Operation.LINK).dict(),
         )
+        icon_url = f"{settings.bugzilla_base_url}/favicon.ico"
         jira_response = self.jira_client.create_or_update_issue_remote_links(
             issue_key=jira_key_in_response,
             link_url=bugzilla_url,
             title=bugzilla_url,
-            icon_url=f"{settings.bugzilla_base_url}/favicon.ico",
+            icon_url=icon_url,
+            icon_title=icon_url,
         )
 
         self.update_issue(payload, bug_obj, jira_key_in_response, is_new=True)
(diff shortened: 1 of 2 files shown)


bsieber-mozilla:
@grahamalama do you think we should move forward with something like this? 

_(If so, we can look into modifying the atlassian-python-api to add the `icon` parameter internal to the remote-issue api call, or moving away from the atlassian-python-api to use the rest api and add the icon later.)_

grahamalama:
Yeah, since the client library doesn't support adding favicons and we're already adding some more context requested in #202 (i.e. `Bugzilla Bug 1234567`), maybe it's best to close this PR and instead:
- "Eject" from `atlassian-python-api` in favor of our own, internal library
or 
- Submit a PR to `atlassian-python-api` to supporting remote link favicons

bsieber-mozilla:
@grahamalama thanks for the addition; I've pointed to the new commit hash to pick those changes up.

grahamalama (approved):
Do you want to set `icon_title` as well?

Also, could you create an issue to revert back to PyPI-released version of `atlassian-python-api` once that patch makes it into the next release?

bsieber-mozilla:
> Do you want to set `icon_title` as well?
When looking at the API and testing this adds a tooltip, but isn't essential (just like `[icon_title](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-remote-links/#api-rest-api-3-issue-issueidorkey-remotelink-post)`). 

I can add it in though.


> Also, could you create an issue to `revert back` to PyPI-released version of `atlassian-python-api` once that patch makes it into the next release?

Currently the toml is pointed to the previous commit when our team last updated the `atlassian-python-api.` 

Good callout; I did not see an issue already made, so I've made a new one for us to migrate to the official release when the new additions are applied.

https://github.com/mozilla/jira-bugzilla-integration/issues/221

PR #216, opened 2022-08-23
Use typed log contexts (fixes #211)

### LINKED ISSUE #211: Create internal schema for objects passed to `extra` parameter of loggers
@grahamalama TODO: fill this issue out

Related: https://mozilla-hub.atlassian.net/browse/SVCSE-707

leplatrem:
Fixes #211

### DIFF
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -417,7 +417,7 @@ class LogContext(BaseModel):
 class RunnerLogContext(LogContext):
     """Logging context from runner"""
 
-    bug = BugId | BugzillaBug
+    bug: BugId | BugzillaBug
     operation: Operation
     action: Optional[Action]
 
@@ -429,3 +429,4 @@ class ActionLogContext(LogContext):
     request: BugzillaWebhookRequest
     jira: JiraContext
     bug: BugzillaBug
+    extra: dict = {}
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -429,4 +429,4 @@ class ActionLogContext(LogContext):
     request: BugzillaWebhookRequest
     jira: JiraContext
     bug: BugzillaBug
-    extra: dict = {}
+    extra: dict[str, str] = {}
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -407,11 +407,9 @@ BugId = TypedDict("BugId", {"id": Optional[int]})
 class LogContext(BaseModel):
     """Generic log context throughout JBI"""
 
-    def set(self, **kwargs):
-        """Set multiple attributes at once"""
-        for k, value in kwargs.items():
-            setattr(self, k, value)
-        return self
+    def update(self, **kwargs):
+        """Return a copy with updated fields."""
+        return self.copy(update=kwargs)
 
 
 class RunnerLogContext(LogContext):
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -9,12 +9,13 @@ import logging
 import warnings
 from inspect import signature
 from types import ModuleType
-from typing import Any, Callable, Literal, Mapping, Optional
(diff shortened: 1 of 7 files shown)


jbi/models.py
```diff
@@ -391,3 +392,41 @@ class BugzillaApiResponse(BaseModel):
...
+
+
+class LogContext(BaseModel):
+    """Generic log context throughout JBI"""
+
+    def set(self, **kwargs):
+        """Set multiple attributes at once"""
+        for k, value in kwargs.items():
+            setattr(self, k, value)
+        return self
```
grahamalama, line 414: I'm seeing that another pattern for updating a model is to use a [model's `copy` method](https://pydantic-docs.helpmanual.io/usage/exporting_models/#modelcopy) with an `update` kwarg:
```python
log_context = log_context.copy(update={action: action})
```
I like that this is a more "models are immutable" approach, and it also re-validates the data.

If you want to stick with the approach you have here, could we:
- maybe rename the method `update`? Visually, `set` with a bunch of kwargs says "initialization", but maybe that's just me.
- validate the model after we update the data?

jbi/runner.py
```diff
@@ -26,28 +26,29 @@ def execute_action(
 
     The value returned by the action call is returned.
     """
-    log_context = {
-        "bug": {
+    log_context = RunnerLogContext(
+        bug={
             "id": request.bug.id if request.bug else None,
         },
```
grahamalama, line 32: Is there _actually_ a case where `request.bug` would be `None`? 

Like in `BugzillaWebhookRequest`, I think we could change bug: `Optional[BugzillaBug]` to `bug: BugzillaBug`. Then we wouldn't need to check for `request.bug`, then we wouldn't need the `BugId` type.

Or maybe if we want to keep the guard(s), we do:

```diff
diff --git a/jbi/runner.py b/jbi/runner.py
index c5a83e8..2d36ef7 100644
--- a/jbi/runner.py
+++ b/jbi/runner.py
@@ -27,9 +27,6 @@ def execute_action(
     The value returned by the action call is returned.
     """
     log_context = RunnerLogContext(
-        bug={
-            "id": request.bug.id if request.bug else None,
-        },
         request=request,
         operation=Operation.HANDLE,
     )
@@ -40,7 +37,7 @@ def execute_action(
         )
         if not request.bug:
             raise IgnoreInvalidRequestError("no bug data received")
-
+        log_context.set(bug=request.bug.dict())
         try:
             bug_obj: BugzillaBug = request.bugzilla_object
         except Exception as ex:

```

jbi/models.py
leplatrem: Oh! I promise I looked for something like this !

jbi/runner.py
leplatrem: Maybe we could return a 400 if no bug is specified instead of going that far and executing the runner... but I feel that this could be another issue/PR

leplatrem: https://github.com/mozilla/jira-bugzilla-integration/issues/218

jbi/runner.py
```diff
@@ -26,28 +26,29 @@ def execute_action(
...
         try:
             bug_obj: BugzillaBug = request.bugzilla_object
         except Exception as ex:
-            logger.exception("Failed to get bug: %s", ex, extra=log_context)
+            logger.exception("Failed to get bug: %s", ex, extra=log_context.dict())
             raise IgnoreInvalidRequestError(
                 "bug not accessible or bugzilla down"
             ) from ex
-        log_context["bug"] = bug_obj.dict()
+        log_context.update(bug=bug_obj)
```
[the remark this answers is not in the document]
leplatrem: well caught!

grahamalama (approved):
🚢

PR #217, opened 2022-08-24
Implement Bugzilla REST API client (fixes #189, fixes #198)

### LINKED ISSUE #189: Replace `python-bugzilla` package with native REST calls
[Slack reference](https://mozilla.slack.com/archives/C016BTJKM9B/p1659540153009049)

### LINKED ISSUE #198: Force Bugzilla client to use REST API
Based on conversation in Slack, we are discouraged from using `xmlrpc` to interact with our Bugzilla instance. The client we're using [uses `xmlrpc` by default](https://github.com/python-bugzilla/python-bugzilla/blob/cd701f6879e7868c0b6b4b37857994b01b88eda1/bugzilla/base.py#L177), but has the option to use the REST API.

Though we have #189 in the backlog, depending on the timeline of that issue, we may want to consider setting `force_rest=TRUE` when using this client, before it's decommissioned.

leplatrem:
fixes #189
fixes #198

* [x] Wait for #220 to be merged
* [x] Add tests using `responses`

### DIFF
--- a/jbi/services.py
+++ b/jbi/services.py
@@ -6,13 +6,13 @@ import logging
 from typing import TYPE_CHECKING
 
 import backoff
-import bugzilla as rh_bugzilla
+import requests
 from atlassian import Jira, errors
 from pydantic import parse_obj_as
 from statsd.defaults.env import statsd
 
 from jbi import environment
-from jbi.models import BugzillaBug, BugzillaComment
+from jbi.models import BugzillaApiResponse, BugzillaBug, BugzillaComment
 
 if TYPE_CHECKING:
     from jbi.models import Actions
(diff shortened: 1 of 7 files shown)


jbi/models.py
```diff
@@ -371,19 +380,17 @@ def map_as_comments(
...
 
-    def getbug_as_bugzilla_object(self) -> BugzillaBug:
-        """Helper method to get up to date bug data from Request.bug.id in BugzillaBug format"""
-        current_bug_info = get_bugzilla().getbug(self.bug.id)  # type: ignore
-        return BugzillaBug.parse_obj(current_bug_info.__dict__)
-
     @functools.cached_property
     def bugzilla_object(self) -> BugzillaBug:
         """Returns the bugzilla bug object, querying the API as needed for private bugs"""
+        from jbi.services import get_bugzilla  # pylint: disable=import-outside-toplevel
```
grahamalama, line 386: This seems like more of a "structural" problem that we could address.

What if instead of instead of `bugzilla_object` and `map_as_jira_comment` being object object methods, they were functions in the `services` module instead?

leplatrem: I agree.

And I think I want to remove the `bugzilla_object` property, since it's redundant with the `bug` attribute. Especially if we land #219.

Instead, I would propose something like this in `runner.py`:

```py
        try:
            request = request.copy(update={"bug": bugzilla_fetch_private(request.bug)})
        except Exception as ex:
            logger.exception("Failed to get bug: %s", ex, extra=log_context)
                "bug not accessible or bugzilla down"
            ) from ex
```

where `bugzilla_fetch_private()` would be:

```py
def bugzilla_fetch_private(bug: BugzillaBug) -> BugzillaBug:
    """Fetch bug private information from API"""
    client = get_bugzilla()
    # If bug is private, fetch all attributes.
    if bug.is_private:
        bug = client.get_bug(bug.id)
    # If comment is private, then webhook does not have comment, fetch it from server
    if bug.comment.is_private:
        comment_list = client.get_comments(bug.id)
        matching_comments = [c for c in comment_list if c.id == bug.comment.id]
        if len(matching_comments) != 1:
            bug.comment = None
        bug.comment = matching_comments[0]
    return bug
```

This way, we have `request.bug` with the proper bug object, and `request.bug.comment` with the proper comment object.

What do you think?

grahamalama: I like it!

leplatrem: I will do it in https://github.com/mozilla/jira-bugzilla-integration/pull/220

bsieber-mozilla (approved):
LGTM

jbi/services.py
```diff
@@ -86,25 +86,51 @@ def jira_visible_projects(jira=None) -> list[dict]:
...
-    def __init__(self, base_url: str, api_key: str):
-        """Constructor"""
-        self._client = rh_bugzilla.Bugzilla(base_url, api_key=api_key)
+    def __init__(self, base_url, api_key):
+        """Initialize the client, without network activity."""
+        self.base_url = base_url
+        self.params = {
+            "api_key": api_key,
+        }
+        self._client = requests.Session()
```
grahamalama, line 102: I don't know if this is a better, worse, or neutral change, but we could also do:
```python
    def __init__(self, base_url, api_key):
        """Initialize the client, without network activity."""
        self.base_url = base_url
        _client = requests.Session()
        _client.params = {"api_key": api_key}
        _client.hooks["response"].append(raise_for_status_hook)
        self._client = _client
```

Where `raise_for_status_hook` is 
```python
def raise_for_status_hook(resp: requests.Response, *args, **kwargs):
    resp.raise_for_status()
```

Then `_call` would just be responsible for parsing JSON and error handling:
```python
    def _call(self, verb, url, *args, **kwargs):
        resp = self._client.request(verb, url, *args, **kwargs)
        parsed = resp.json()
        if parsed.get("error"):
            raise BugzillaClientError(parsed["message"])
        return parsed
```

jbi/services.py
```diff
@@ -6,13 +6,13 @@
 from typing import TYPE_CHECKING
 
 import backoff
-import bugzilla as rh_bugzilla
+import requests
```
grahamalama, line 9: I notice that we don't have `requests` in `pyproject.toml`. Should we add it?

leplatrem: Yes!

jbi/services.py
leplatrem: I think I prefer having everything in `_call()`. It's not like we have `request.get()` calls everywhere in the code base, the hook approach does not bring much here.
I removed the `self.params` class member to make `_call()` even more confined

PR #219, opened 2022-08-25
Return 4XX response if no bug is provided (fixes #418)

leplatrem:
I have checked logs, and we haven't logged `no bug data received` a single time in the last 7 days. So, I would conclude that this is not a case we want to support (eg. return a gentle 200 if no bug provided).

Screenshot 2022-08-25 at 12 11 35

Unlike `no action matching bug whiteboard tags` for example...

Screenshot 2022-08-25 at 11 56 32

### DIFF
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -317,7 +317,7 @@ class BugzillaWebhookRequest(BaseModel):
     webhook_id: int
     webhook_name: str
     event: BugzillaWebhookEvent
-    bug: Optional[BugzillaBug]
+    bug: BugzillaBug
 
     def map_as_jira_comment(self):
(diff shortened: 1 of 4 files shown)


bsieber-mozilla (approved):
Thanks for this update! If the data isn't there, ValidationError and fail. 
👍

bsieber-mozilla:
@leplatrem I noticed there was a conflict with `main` on this branch; I resolved the merge conflict and force-pushed the branch here: #219 

Feel free to merge that in or resolve in branch (_I didn't force-push the merge resolution on your branch because I didn't want to cause confusion/git conflicts.)_

leplatrem:
Thanks I merged it 🙏

PR #220, opened 2022-08-25
Reorganize calls to Bugzilla into a dedicated class in `services.py`

leplatrem:
* [x] Refactor
* [x] Remove useless assignments in tests
* [x] Remove notion of private bugs in tests (now responsibility of runner)
* [x] Wait for #219 to clean up cases where `payload.bug` can be `None`

### DIFF
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -315,13 +315,6 @@ class BugzillaBug(BaseModel):
 class BugzillaWebhookRequest(BaseModel):
     """Bugzilla Webhook Request Object"""
 
-    class Config:
-        """pydantic model config"""
-
-        keep_untouched = (
-            functools.cached_property,
-        )  # https://github.com/samuelcolvin/pydantic/issues/1241
-
     webhook_id: int
     webhook_name: str
     event: BugzillaWebhookEvent
--- a/jbi/models.py
+++ b/jbi/models.py
@@ -25,7 +25,6 @@ from pydantic_yaml import YamlModel
 
 from jbi import Operation
 from jbi.errors import ActionNotFoundError
-from jbi.services import get_bugzilla
 
 logger = logging.getLogger(__name__)
 
@@ -330,20 +329,11 @@ class BugzillaWebhookRequest(BaseModel):
 
     def map_as_jira_comment(self):
         """Extract comment from Webhook Event"""
-        comment: BugzillaWebhookComment = self.bug.comment
         commenter: BugzillaWebhookUser = self.event.user
-        comment_body: str = comment.body
-
-        if comment.is_private:
-            bug_comments = get_bugzilla().get_comments([self.bug.id])
-            comment_list = bug_comments["bugs"][str(self.bug.id)]["comments"]
-            matching_comments = [c for c in comment_list if c["id"] == comment.id]
-            if len(matching_comments) != 1:
-                return None
-            comment_body = matching_comments[0]["text"]
-
-        body = f"*({commenter.login})* commented: \n{{quote}}{comment_body}{{quote}}"
-        return body
+        comment: BugzillaWebhookComment = self.bug.comment
+        if comment is None:
+            return None
+        return f"*({commenter.login})* commented: \n{{quote}}{comment.body}{{quote}}"
 
     def map_as_jira_description(self):
         """Extract description as comment from Webhook Event"""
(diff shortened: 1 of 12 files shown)


jbi/runner.py
```diff
@@ -38,23 +39,26 @@ def execute_action(
         )
 
         try:
-            bug_obj: BugzillaBug = request.bugzilla_object
+            if request.bug.is_private:
+                request = request.copy(
```
grahamalama, line 43: One thing I'm concerned with here: while it's more convenient to set `request.bug` as the result of `get_bugzilla().get_bug`, I'm wondering if that will muddle our logging if over the course of an action run, we modify the request. (e.g., when we received the request it was a private bug, but now it's the fetched bug with all of the data. How did that happen 🤔)

Maybe we should modify `__call__` signatures to that they accept `bug` and `event` as parameters. Then we can do:

```python
bug = request.bug
if request.bug.is_private:
    bug = get_bugzilla().get_bug(bug.id)
handled, details = action.caller(bug=bug, event=request.event)
```

Or maybe that's logic we encourage action authors to handle _in_ call:

```python
def __call__(self, payload):
    bug, event = payload.bug, payload.event
    if bug.is_private:
        bug = get_bugzilla.get_bug(bug.id)
    ...
```

but in both cases, it feels to me like the request should be treated as immutable. 
What do you think?

jbi/runner.py
```diff
@@ -38,23 +39,26 @@ def execute_action(
...
+                    update={"bug": get_bugzilla().get_bug(request.bug.id)}
```
grahamalama, line 44: This pattern seems handy enough that maybe the method you added to `LogContext` should be added to a `BaseModel` that all of our models. Not requesting a change here, just thinking out loud.

jbi/models.py
```diff
@@ -330,20 +329,11 @@ class Config:
...
-            if len(matching_comments) != 1:
-                return None
-            comment_body = matching_comments[0]["text"]
-
-        body = f"*({commenter.login})* commented: \n{{quote}}{comment_body}{{quote}}"
-        return body
+        comment: BugzillaWebhookComment = self.bug.comment
+        if comment is None:
+            return None
+        return f"*({commenter.login})* commented: \n{{quote}}{comment.body}{{quote}}"
```
grahamalama, line 336: Do not need to account for private comments anymore? I wonder if we should raise a `ValueError` if `comment.is_private`.

jbi/services.py
```diff
@@ -84,15 +85,53 @@ def jira_visible_projects(jira=None) -> list[dict]:
...
+            matching_comments = [c for c in comment_list if c.id == bug.comment.id]
+            found = matching_comments[0] if matching_comments else None
+            bug = bug.copy(update={"comment": found})
+        return bug
+
+    def get_comments(self, bugid) -> list[BugzillaComment]:
+        """Return the list of comments for the specified bug ID"""
+        response = self._client.get_comments(idlist=[bugid])
+        comments = response["bugs"][str(bugid)]["comments"]
+        return [BugzillaComment.parse_obj(comment) for comment in comments]
```
grahamalama, line 118: Could also use [parse_obj_as](https://pydantic-docs.helpmanual.io/usage/models/#parsing-data-into-a-specified-type):

```python
return parse_obj_as(list[BugzillaComment], comments)
```

tests/conftest.py
```diff
@@ -32,9 +32,12 @@ def settings():
...
 @pytest.fixture(autouse=True)
-def mocked_bugzilla():
-    with mock.patch("jbi.services.rh_bugzilla.Bugzilla") as mocked_bz:
-        yield mocked_bz()
+def mocked_bugzilla(request):
+    if "no_mocked_bugzilla" in request.keywords:
+        yield None
+    else:
+        with mock.patch("jbi.services.BugzillaClient") as mocked_bz:
+            yield mocked_bz()
```
grahamalama, line 40: Either this, or remove `autouse=True` and explicitly include `mocked_bugzilla` where needed.

jbi/runner.py
bsieber-mozilla: I agree in the long term modifying the request could be confusing.

I think between the suggestions, I prefer the latter; providing the 'payload' or a 'data' object to the action.

(When we discuss composition of actions, the generic 'data' object could be part of a consistent method_signature for internal tasks that could be ordered and enabled, breaking Actions down into smaller tasks?)

But also, not sure if this should be tackled in this PR.

bsieber-mozilla (approved):
Thanks for this change!

tests/conftest.py
leplatrem: I went with this route, since we only want to exclude in `test_services.py` where we test its internals

jbi/runner.py
leplatrem: I agree, removing mutations of the original request is better.

I kind of prefer passing the `bug` and `event` objects to the actions, instead of duplicating the extraction in the actions. Even if that means we potentially have an additional fetch for actions that may not require it.

If you don't mind, I'd like to address this in a follow-up, so that we can ship #224 and #217 first

jbi/models.py
leplatrem: Comment is `None` if comment is private. I'll add a comment, but I didn't want to change the previous behavior

jbi/runner.py
grahamalama: Can we open an issue to track this?

jbi/models.py
grahamalama: I'm all for returning `None`, but I was wondering if we still needed the guard. What happens if we try to access `comment.body` on a private or nonexistent `webhookrequest.bug.comment`?

jbi/runner.py
leplatrem: https://github.com/mozilla/jira-bugzilla-integration/issues/227

leplatrem:
I'm adding the `breaking-change` label, because we renamed the metric `get_bug`

PR #223, opened 2022-08-26
Fix extract_from_see_also() with None values

leplatrem:


jbi/runner.py
```diff
@@ -27,8 +27,8 @@ def execute_action(
     The value returned by the action call is returned.
     """
     log_context = RunnerLogContext(
-        bug=request.bug.dict(),
-        request=request.dict(),
+        bug=request.bug,
+        request=request,
```
leplatrem, line 31: drive by 🙄

grahamalama (approved):

bsieber-mozilla (approved):

jbi/runner.py
bsieber-mozilla: Accident?

bsieber-mozilla (approved):
The check looks to be failing because of no Label, but there is a 'bug' label applied. The change looks good to me.

jbi/runner.py
leplatrem: 😄  Somewhere between #216 and #222/#219 since the PRs were done simultaneously

PR #224, opened 2022-08-26
Separate and "cache" services

grahamalama:
This PR:
- separates the Bugzilla and Jira clients + functions in separate modules
- "caches" the services by making them singletons


leplatrem:
Ouch, this conflicts a lot with #217 :( I can give you a hand next week

grahamalama:
> Ouch, this conflicts a lot with #217 :( I can give you a hand next week

Yeah I figured that'd happen -- no worries. Hopefully it'll mostly be a matter of swapping out the `jbi.services.bugzilla` module with the new client you've written

jbi/services/bugzilla.py
```diff
@@ -0,0 +1,128 @@
...
+        parsed = BugzillaApiResponse.parse_obj(updated_info)
+        if not parsed.bugs:
+            raise BugzillaClientError(
+                f"Unexpected response content from 'PUT {url}' (no 'bugs' field)"
+            )
+        return parsed.bugs[0]
+
+
+instrumented_methods = (
+    "getbug",
```
leplatrem, line 97: ⚠️ I renamed it to `get_bug` in #220

jbi/router.py
```diff
@@ -37,8 +37,11 @@ def root(request: Request, settings: Settings = Depends(get_settings)):
 @router.head("/__heartbeat__")
 def heartbeat(response: Response, actions: Actions = Depends(get_actions)):
     """Return status of backing services, as required by Dockerflow."""
-    health_map = jbi_service_health_map(actions)
-    health_checks = []
+    health_map = {
+        "bugzilla": bugzilla.check_health(),
+        "jira": jira.check_health(actions),
```
leplatrem, line 42: So, we only leave `check_health` at the module level. Would it make more sense to have it as a client method to align it with the rest of the actions code? eg. `jira.get_client().check_health(actions)` ?

jbi/services/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Exports the bugzilla and jira services modules for convenience"""
+
+from . import bugzilla, jira
```
leplatrem, line 3: You almost could have left `get_jira` and `get_bugzilla` here.

```py
_BZ_CLIENT = None

def get_bugzilla(*args, **kwargs):
    global _CLIENT  # pylint: disable=global-statement
    if not _BZ_CLIENT:
        _BZ_CLIENT = bugzilla.get_client()
    return _BZ_CLIENT
```

(Just a drive-by idea, not a change request)

jbi/services/bugzilla.py
```diff
@@ -0,0 +1,128 @@
...
+    "update_bugs",
+)
+
+_CLIENT = None
+
+
+def get_client():
+    """Get bugzilla service"""
+    global _CLIENT  # pylint: disable=global-statement
+    if not _CLIENT:
```
leplatrem, line 108: Wouldn't have `functools.lru_cache()` have worked here?

tests/conftest.py
```diff
@@ -33,17 +33,19 @@ def settings():
 
 
 @pytest.fixture(autouse=True)
-def mocked_bugzilla(request):
+def mocked_bugzilla(request, monkeypatch):
     if "no_mocked_bugzilla" in request.keywords:
         yield None
     else:
-        with mock.patch("jbi.services.BugzillaClient") as mocked_bz:
+        monkeypatch.setattr("jbi.services.bugzilla._CLIENT", None)
```
leplatrem, line 40: You didn't use lru_cache because you needed a way to mock? 
I think you could call `cache_clear()` on a lru_cache decorated function

jbi/router.py
```diff
@@ -87,7 +90,7 @@ def get_whiteboard_tags(
 @router.get("/jira_projects/")
 def get_jira_projects():
     """API for viewing projects that are currently accessible by API"""
-    visible_projects: list[dict] = jira_visible_projects()
+    visible_projects: list[dict] = jira.fetch_visible_projects()
```
leplatrem, line 93: Do you think we should have our own Jira class with such methods? Or can we a general guideline on when to use a module function and when to use a client method?

jbi/runner.py
```diff
@@ -9,7 +9,7 @@
 from jbi.environment import Settings
 from jbi.errors import ActionNotFoundError, IgnoreInvalidRequestError
 from jbi.models import Actions, BugzillaWebhookRequest, RunnerLogContext
-from jbi.services import get_bugzilla
+from jbi.services import bugzilla
```
leplatrem, line 12:
```suggestion
from jbi.services.bugzilla import get_client as get_bugzilla
```
in the sake of brevity?

(`get_client()` is also more explicit, so...)

leplatrem (approved):
Please fix the metric name, otherwise it looks good!
Feel free to ignore the other comments if you don't find them relevant

jbi/services/bugzilla.py
grahamalama: I think I avoided `lru_cache` at first because the Redhat Bugzilla client had side effects on instantiation, but now that we have our own client that doesn't do this and the Jira client also doesn't have any side effects on instantiation, `lru_cache` does seem nicer.

tests/conftest.py
grahamalama: Good call, thanks!

jbi/services/bugzilla.py
grahamalama: Good catch. Fixed in [3823706](https://github.com/mozilla/jira-bugzilla-integration/pull/224/commits/3823706f0d875c726bed7c5ea0cf1b85d3a8671c)

jbi/router.py
grahamalama: > Or can we establish a general guideline on when to use a module function and when to use a client method

I like this approach. I was thinking about this a bit. I think the separation of responsibilities should be [^1]:
- client class: calls the REST API, returns JSON
- module functions or a service class: uses the client to fetch data, then processes it

If an action wants to just fetch `json`, it could do that with the REST client. But I'd imagine they'd more than likely use a module function or service class method.

> Do you think we should have our own Jira class with such methods?

For now, I don't think the service module functions share any state that would make combining them into a class useful. But if that starts to become the case, maybe a service class makes sense.

[^1]: I know this isn't how the Bugzilla module is at the moment, but maybe we could make that change?

jbi/runner.py
grahamalama: Yeah, if we wanted to include the `client` context in that import alias, I think it'd be
`get_bugzilla_client` vs `bugzilla.get_client`, so same # of characters

jbi/router.py
grahamalama: I think this comment kind of goes with `https://github.com/mozilla/jira-bugzilla-integration/pull/224#discussion_r958466469`? Maybe we should have a design discussion about classes / clients/ method access and make a change outside this PR.

jbi/services/__init__.py
grahamalama: Yeah, I kept it this way for now since we have `<service module>.get_client` and `<service module>.other_function`. If we end up putting service access through one class, then getters in `init` probably make more sense

PR #226, opened 2022-08-26
Update deps to remove warning

grahamalama:
When running `poetry install --no-root` locally, I noticed a warning:

> Warning: The lock file is not up to date with the latest changes in
> pyproject.toml. You may be getting outdated dependencies. Run update
> to update them.

Instead of running that command though, I ran `poetry lock --no-update`
to regenerate the lockfile. Then I ran `poetry install --no-root` again,
which resulted in the following updates:

> Updating coverage (6.4.3 -> 6.4.4)
> Updating uvicorn (0.18.2 -> 0.18.3)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -11,9 +11,9 @@ idna = ">=2.8"
 sniffio = ">=1.1"
 
 [package.extras]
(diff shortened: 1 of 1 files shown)


bsieber-mozilla (approved):
LGTM, thanks for updating these deps

PR #230, opened 2022-08-30
Do not mutate the request object (fixes #227)

### LINKED ISSUE #227: Do not mutate the incoming request object
As reported in https://github.com/mozilla/jira-bugzilla-integration/pull/220#pullrequestreview-1087240707

While it's more convenient to set `request.bug` as the result of `get_bugzilla().get_bug`, it could muddle our logging if over the course of an action run, we modify the request. (e.g., when we received the request it was a private bug, but now it's the fetched bug with all of the data. How did that happen 🤔)

Maybe we should modify `__call__` signatures so that they accept `bug` and `event` as parameters. Then we can do:

```python
bug = request.bug
if request.bug.is_private:
    bug = get_bugzilla().get_bug(bug.id)
handled, details = action.caller(bug=bug, event=request.event)
```

Or maybe that's logic we encourage action authors to handle _in_ call:

```python
def __call__(self, payload):
    bug, event = payload.bug, payload.event
    if bug.is_private:
        bug = get_bugzilla.get_bug(bug.id)
    ...
```

but in both cases,  the request should be treated as immutable.

leplatrem:
fixes #227


grahamalama (approved):
🚢

Issue #77, opened 2022-07-07
How to extend existing actions?


leplatrem:
In #75 we are adding more logic to the default action behavior. Namely setting assignee and status fields.

The chosen approach was to extend the default action behavior using inheritance.

**Pros**

- No duplicated code between actions

**Cons**

- Indirection. One has to read and mentally merge two classes in order to understand the final behavior
- Hard to pick (eg. I want status to be set, but not the assignee)

There could be other approaches to consider:

## Single action with (many) parameters

```yaml
    action: src.jbi.whiteboard_actions.default_action
    contact: [example@allizom.com]
    description: example configuration
    enabled: true
    parameters:
      jira_project_key: EXMPL
      whiteboard_tag: example
      sync_assignee: True
      status_map:
        NEW: "In Progress"
        FIXED: "Closed"
```

**Pros**

- A single piece of code to read  (`default_action.py`)

**Cons**

- The default action code could potentially grow indefinitely, and we would loose the extensibility offered by the action API
- Unit tests have to combine a lot of combinations for the parameters

## Support execution of multiple atomic actions

Instead of a single action, define a list of actions to be executed. The specified parameters would be passed to all to simplify configuration.

```yaml
    contact: [example@allizom.com]
    description: example configuration
    enabled: true
    actions:
       - src.jbi.whiteboard_actions.create_or_update
       - src.jbi.whiteboard_actions.comments
       - src.jbi.whiteboard_actions.assignee
       - src.jbi.whiteboard_actions.status
    parameters:
      jira_project_key: EXMPL
      whiteboard_tag: example
      status_map:
        NEW: "In Progress"
        FIXED: "Closed"
```

**Pros**

- Smaller bits of code
- Modularity
- Easier testing

**Cons**

- Potential duplication in config if all use-cases want the same set of actions
- Multiple calls to API for separate fields (verbose history in jira?)

----

There could be other approaches. Maybe the current one is fine, but at least this issue could help us drive a conversation about it (before the code would grow too much)

Mossop:
One of the things I tried first in #75 was to have a couple of methods on the default action that generated the list of fields to update in Jira:

```python
    def fields_for_create(self, payload, bug_obj):
        return {
            **bug_obj.map_as_jira_issue(),  # type: ignore
            "issuetype": {"name": bug_obj.issue_type()},
            "description": comment_list["bugs"][str(bug_obj.id)]["comments"][0]["text"],
[...]

PR #233, opened 2022-09-01
Remove class inheritance between default and default with status and assignee (ref #77)

leplatrem:
By removing the inheritance between the two actions, we pave the way for #77 and the resulting create/update/update workflows, although they have some duplicate bits, can be read from top to bottom.


bsieber-mozilla (approved):
I like the chaining of the "maybe" operations, I think of these as (Conditions+Process) in one method.

Let's land this then continue to iterate on naming and design?

PR #236, opened 2022-09-07
Chain operations instead of using nested conditions (with `:=` operators)

leplatrem:
Summary of changes:

* actions now take an `ActionContext` object, instantiated by the `runner`
* actions chain steps, altering and passing the context object
* the runner is now in charge of determining with operation is going on (create, update, comment)
* the tests fixtures were adapted accordingly

This is a segway in order to split the `default` and `default_with_assignee_status` into atomic steps (currently implemented as `maybe_*()` functions. 

In another PR we could add `jira_client` and `bugzilla_client`  to the context (composition pattern).


leplatrem:
@bsieber-mozilla @grahamalama since #253 was approved before this PR, I'm a bit confused whether I should merge it. What do you think? Do you want to review (by commit) here?

bsieber-mozilla (approved):

PR #242, opened 2022-09-08
Pass the correct fields to add see-also links to the REST API.

Mossop:


Mossop:
For reference the correct form for the `see_also` object is documented here: https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html#update-bug

leplatrem (approved):
I'll add tests in a follow-up

PR #253, opened 2022-09-13
Load default action steps from configuration

leplatrem:
This removes the hard-coded steps from `default.py` and deletes the `default_with_assignee_and_status` action.


README.md
```diff
@@ -125,32 +92,88 @@ If configured, the action supports setting the Jira issues's status when the Bug
...
+
+```yaml
+- whiteboard_tag: fidefe
+  contact: example@allizom.com
+  description: full configuration
+  enabled: true
+  module: jbi.actions.default
+  parameters:
+    jira_project_key: FIDEFE
+    steps:
```
bsieber-mozilla, line 114: It appears that this "steps" section splits based on the event-type (which doesn't seem to be mentioned); adding these event types in the documentation might be helpful.

Another concern is, during the "update" event there is a possibility of creating a jira issue; based on the step name "update_issue" this implies an issue would be already present. Changing the name to "upsert" instead of "update" might work if the create functionality is still there. 


---

However, I question the value of separating these steps by the "events" (create, update, comment, attachment). 

I think each step should validate that if it should/n't run: 
- create_issue runs on both "update" and "create" events, only create a Jira ticket if this was already done (when "comment" events occur, do nothing)

---

Just some minor thoughts

config/config.prod.yaml
```diff
@@ -19,9 +19,25 @@
   contact: gijs@mozilla.com
   description: Firefox Front End whiteboard tag
   enabled: true
-  module: jbi.actions.default_with_assignee_and_status
+  module: jbi.actions.default
```
bsieber-mozilla, line 22:
```suggestion
```
since default is used; can this be removed?

README.md
```diff
@@ -66,56 +66,23 @@ A bit more about the different fields...
 ## Available Actions
 
 ### Default
-The `jbi.actions.default` action will create or update the Jira issue and its comments.
-It will also set the Jira issue URL in the Bugzilla bug `see_also` field.
+The `jbi.actions.default` action will take the list of steps to be executed when
+the Webhook is received from configuration.
+When none is specified, it will create or update the Jira issue, publish comments when
+assignee, status, or resolution are changed, or when a comment is posted on the Bugzilla ticket.
```
grahamalama, line 72: > publish comments when assignee, status, or resolution are changed

Does this actually publish a comment? Or would this fall under "create or update the Jira issue"?

README.md
```diff
@@ -66,56 +66,23 @@ A bit more about the different fields...
...
 
 **Parameters**
 
 - `jira_project_key`
     - string
     - The Jira project identifier
-- `sync_whiteboard_labels`
+- `steps`
+    - mapping [str, list[str]]
+    - If defined, the specified steps are executed. The group of steps listed under `new` are executed when a Bugzilla event occurs on a ticket that is unknown to Jira. The steps under `existing`, when the Bugzilla ticket is already linked to a Jira issue. The steps under `comment` when a comment is posted on a linked Bugzilla ticket.
```
grahamalama, line 84: ✅ Nice.

README.md
```diff
@@ -125,32 +92,88 @@ If configured, the action supports setting the Jira issues's status when the Bug
...
+- `maybe_assign_jira_user`:
+  It will attempt to assign the Jira issue the same person as the bug is assigned to. This relies on
+  the user using the same email address in both Bugzilla and Jira. If the user does not exist in Jira
+  then the assignee is cleared from the Jira issue. The Jira account that JBI uses requires the "Browse
+  users and groups" global permission in order to set the assignee.
+- `maybe_update_issue_resolution`:
+  If the Bugzilla ticket resolution field is specified in the `resolution_map` parameter, it will set the
+  Jira issue resolution.
+- `maybe_update_issue_status`:
+  If the Bugzilla ticket status field is specified in the `status_map` parameter, it will set the
```
grahamalama, line 165: These explanations seem like they'd be valuable to include in the function docstrings as well

config/config.nonprod.yaml
```diff
@@ -12,9 +12,24 @@
...
+        - maybe_assign_jira_user
+        - maybe_update_issue_resolution
+        - maybe_update_issue_status
+      existing:
+        - update_issue
+        - maybe_assign_jira_user
+        - maybe_update_issue_resolution
+        - maybe_update_issue_status
+      comment:
+        - create_comment
```
grahamalama, line 32: Is there a way to set up the configuration where we only need to override the step section that deviates from default? Like in this "custom" workflow and the default workflow, the comment steps remain the same, so it might be nice to not have to specify that.

jbi/actions/steps.py
```diff
@@ -1,89 +1,92 @@
...
+from jbi import Operation
 from jbi.models import ActionContext
-from jbi.services import jira
+from jbi.services import bugzilla, jira
 
 logger = logging.getLogger(__name__)
 
 
-JIRA_REQUIRED_PERMISSIONS = DEFAULT_JIRA_REQUIRED_PERMISSIONS
+def create_comment(context: ActionContext, **parameters):
```
grahamalama, line 16: I wonder if we should create a type for this function signature and enforce it somewhere. Probably not something for this PR, but something I'm thinking of.

grahamalama (commented):
Looking good! A first round of comments.

README.md
leplatrem: With the default action, JBI posts a comment on the Jira when assignee, status, or resolution is changed on the Bugzilla ticket.

With the default_with_assignee_and... one, when these fields change on Bugzilla, JBI set them on the Jira issue, but does not post any comment.

config/config.nonprod.yaml
leplatrem: Do you mean like a merge of default steps and the configuration ones?

For example, if the `comment` group is not specified we use the default steps for `comment`. If it is specified we follow the configuration (eg. specify it and leave it empty to remove the default behaviour)

leplatrem: I did the merge, let me know if that's what you had in mind

grahamalama: Yep, that's what I was talking about! Nice ✅

README.md
grahamalama: Ahh interesting. Thanks for the clarification.

jbi/actions/steps.py
grahamalama: https://github.com/mozilla/jira-bugzilla-integration/issues/262

grahamalama (approved):
🚀

PR #276, opened 2022-10-06
 Fix #240: log responses objects 

### LINKED ISSUE #240: Log API HTTP responses obtained from Jira and Bugzilla
Currently we return the responses in the webhook endpoint, but we don't log them. Hence, we can't properly troubleshoot what is going on.

leplatrem:


bsieber-mozilla (approved):
LGTM; getting more visibility into the response is helpful--thanks!

PR #277, opened 2022-10-06
Align versions. Use PYPROJECT.TOML as main version.

bsieber-mozilla:


Makefile
```diff
@@ -18,6 +18,7 @@ help:
 	@echo "  clean         - clean local cache folders"
 	@echo "  format        - run formatters (black, isort), fix in place"
 	@echo "  lint          - run linters"
+	@echo "  version-bump  - bump version (major|minor|patch)"
```
bsieber-mozilla, line 21:
```suggestion
	@echo "  version - print version or provide PARAM (major|minor|patch) to bump version"
```

leplatrem (approved):
I like `make version`  or `make bump-version`. 
If we could find a better name than `PARAM=`, it would be great though ;)
Or just `make bump-patch`, `make bump-minor`, `make bump-major` ?

bsieber-mozilla:
> If we could find a better name than `PARAM=`, it would be great though ;) Or just `make bump-patch`, `make bump-minor`, `make bump-major` ?

Yeah, in terms of MAKEFILE's I'm curious what's the better approach... Maybe providing both the parameterized one with (SEMVAR/VERSION instead of PARAM) and the provide the other bump methods for convenience. There's other options beyond patch, minor, major--prefixed with pre--as well as prerelease.
'''
VERSION=major make bump-version
make bump-patch
make bump-minor
make bump-major
'''

I hadn't used MAKE in a parameterized way like this before--not sure if that's an anti-pattern, or just a pattern we want to avoid...

bsieber-mozilla:
_Missing piece: Git versioning_

leplatrem:
According to our conversation elsewhere, an alternative could be:
* Rely on Git tags only
* Read version from `version.json` in app (generated from git tag during container build)
* Remove version (set to 0) from pyproject.toml

bsieber-mozilla:
#291 merged; closing this

PR #325, opened 2022-11-21
[FXDROID] updating status_map based on available status workflow on j…

bsieber-mozilla:
…ira board.


bsieber-mozilla:
Manually went into nonprod and prod jira instances to confirm available workflow/transitions of status on jira boards. 

Perhaps this is something that can be automated/tested prior to push--and present available status options?

PR #359, opened 2023-01-10
Flip default config value `enabled` to `true`

grahamalama:
One of the available project config values is `enabled`. This value controls whether actions are run when a matching webhook is encountered. At the time of this commit, each project sets `enabled: true`. So, in this commit, we flip the default so that actions are enabled by default and can optionally be disabled by setting `enabled: false`.


bsieber-mozilla (approved):
LGTM, thanks @grahamalama!

PR #363, opened 2023-01-12
Ignore `__lbheartbeat__` requests in Sentry

### LINKED ISSUE #215: Do we need so many Sentry transactions?
In the week of 15 to 22nd of August, the hits on the `__lbheartbeat__`  endpoint led to 2.6 millions Sentry transactions.
Screenshot 2022-08-23 at 11 24 26

Do we need them? Do they cost anything? Should we disable performance reports?

grahamalama:
Closes #215

In this PR, we:
- use dynamic sampling to ignore `__lbheartbeat__` requests in Sentry. The `sentry_traces_sample_rate`environment variable is still present so that we can control the global sample rate of non-`__lbheartbeat__` requests.
- update the way we initialize Sentry to be in line with current [documentation](https://docs.sentry.io/platforms/python/guides/fastapi/)
- rename `test_main` to `test_app`. `main.py` was renamed to `app.py`, but we didn't rename the complementary test module at that time


jbi/app.py
```diff
@@ -21,13 +19,20 @@
...
+
+
 sentry_sdk.init(
     dsn=settings.sentry_dsn,
-    integrations=[
-        StarletteIntegration(),
-        FastApiIntegration(),
-    ],
-    traces_sample_rate=settings.sentry_traces_sample_rate,
+    traces_sampler=traces_sampler,
```
bsieber-mozilla, line 35: I think I prefer the  defaulted "url" vs the "endpoint"; so nothing to change here--if we wanted to modify the transaction name presented in sentry in the future, we could look to this anchor (of the link you provided) https://docs.sentry.io/platforms/python/guides/fastapi/#integration-options

bsieber-mozilla (approved):
LGTM! 🚀

jbi/app.py
bsieber-mozilla: _(resolving this comment to unblock)_

Issue #21, opened 2022-03-14
Setting up Webhooks with 'ANY' Product


bsieber-mozilla:
Tracking this issue in bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1752191

leplatrem:
With this in place, the product filtering would be done in JBI, and only one webhook would have to be configured on Bugzilla (instead of 1 per product). 
This is not a requirement but would make operations easier

bsieber-mozilla:
@leplatrem et al, I think we can confidently enable the webhooks with 'ANY' for production when we've established a bit more fidelity around observability--such as alerts, dashboards, and enhanced logging...

required~ before we make this change: #181,  #82, #193 
suggested before we make this change: #140, #189, #58

Currently we've got a bugzilla webhook queue per-project, meaning that when the flow of a particular project is disabled, the webhooks tracking other projects can still run; when we convert to "ANY" we'll have a single queue in bugzilla and single point of operations.

bsieber-mozilla:
_Why does the "ANY" product webhook get automatically disabled?_

bsieber-mozilla:
```Product: ANY
Component: ANY```
--now functioning... required disabling all existing webhooks and creating a new one (toggling ON/OFF the webhook with ANY resulted in internal server errors for BMO).

When we make the shift from N Products-> ANY Products; this will likely result in a temporary down time as we disable all current, and enable the new "ANY" webhook.

<img width="1215" alt="Screenshot 2023-02-08 at 1 33 39 PM" src="https://user-images.githubusercontent.com/53873986/217656050-30480b41-9260-4aeb-a24f-6c981ed25227.png">

bsieber-mozilla:
> required~ before we make this change: #181, #82, #193
> suggested before we make this change: #140, #189, #58

It looks like of this list:
#181 and #58 remain; the latter is not a prerequisite 

#181 is required since the amount of data is now increased and the error queue is down to a single queue--instead of per-product. This means when JBi experiences an issue in any of it's configurations--that can impact all data-flows; #390 would assist/mitigate this issue

#181 and #390 will be valuable for us to safely migrate to using "ANY" in production

bsieber-mozilla:
Yesterday the JBI-PROD-Toolkit Queue went upwards of 100+ errors--remaining queues were chugging away at events.

[...]

PR #395, opened 2023-02-09
Fix #181: check webhooks status in heartbeat

### LINKED ISSUE #181: Can we ping Bugzilla to see if the webhook is enabled?
We've experienced a few system disruptions caused by Bugzilla disabling the webhook that we've set up to push events to our service.

Is there a way for us to send some request to Bugzilla to check on the status of the webhook? If so, this is something we could potentially incorporate into the healthcheck endpoint.

leplatrem:
Fix #181

* [x] Basic functionnality tested manually on bugzilla-dev.allizom.org
* [x] Unit tests
* [x] Decide what to do with #21 (either clean up disabled hooks or modify heartbeat logics) 
* [x] Wait for deploy of https://bugzilla.mozilla.org/show_bug.cgi?id=1800750 in PROD. See https://bugzilla.mozilla.org/__version__

### DIFF
--- a/jbi/services/bugzilla.py
+++ b/jbi/services/bugzilla.py
@@ -9,7 +9,13 @@ import requests
 from pydantic import parse_obj_as
 
 from jbi import Operation, environment
-from jbi.models import ActionContext, BugzillaApiResponse, BugzillaBug, BugzillaComment
+from jbi.models import (
+    ActionContext,
+    BugzillaApiResponse,
+    BugzillaBug,
+    BugzillaComment,
+    BugzillaWebhooksResponse,
+)
 
 from .common import InstrumentedClient, ServiceHealth
 
@@ -95,6 +101,17 @@ class BugzillaClient:
             )
         return parsed.bugs[0]
 
+    def list_webhooks(self):
+        """List the currently configured webhooks, including their status."""
+        url = f"{self.base_url}/rest/webhooks/list"
+        webhooks_info = self._call("GET", url)
+        parsed = BugzillaWebhooksResponse.parse_obj(webhooks_info)
+        if parsed.webhooks is None:
+            raise BugzillaClientError(
+                f"Unexpected response content from 'GET {url}' (no 'webhooks' field)"
+            )
+        return parsed.webhooks
+
 
 @lru_cache(maxsize=1)
 def get_client():
@@ -109,6 +126,7 @@ def get_client():
             "get_bug",
             "get_comments",
             "update_bugs",
+            "list_webhooks",
         ),
         exceptions=(
             BugzillaClientError,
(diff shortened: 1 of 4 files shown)


bsieber-mozilla:
https://bmo.readthedocs.io/en/latest/extensions/Webhooks/api/v1/webhooks.html

_API + Docs are live in the latest build_

jbi/services/bugzilla.py
```diff
@@ -120,7 +138,28 @@ def get_client():
...
+        for webhook in jbi_webhooks:
+            if not webhook.enabled:
+                all_webhooks_enabled = False
+                logger.error(
+                    "Webhook %s is disabled (%s errors)", webhook.name, webhook.errors
+                )
+
+    health: ServiceHealth = {
+        "up": logged_in,
+        "all_webhooks_enabled": all_webhooks_enabled,
```
bsieber-mozilla, line 161: If this is unhealthy; how do we notify about the error? _(sentry? logs? other?)_

bsieber-mozilla (approved):
Thanks for this PR!

jbi/services/bugzilla.py
bsieber-mozilla: I think "errors" can be >0 for a substantial amount of time before the queue is "disabled" but the amount of errors would keep increasing.

I think both "disabled" and "errors" would be worthwhile to track for healthiness

leplatrem: > If this is unhealthy; how do we notify about the error? (sentry? logs? other?)

In this case the `logger.error()` would send a Sentry error.

> I think both "disabled" and "errors" would be worthwhile to track for healthiness

Should we emit a warning when errors are `>0` then?

bsieber-mozilla: > In this case the logger.error() would send a Sentry error.
👍 

> Should we emit a warning when errors are >0 then?
Yeah, a threshold of 0 sounds good.

bsieber-mozilla:
> @bsieber-mozilla Are we already using the `product=Any` webhook? Were all the former webhooks removed? Because if they are still there and disabled the heartbeat will fail. I'm not sure we want to be smart about this, like detect that there is an `Any` webhook enabled, and thus ignore all the disabled ones. Let me know what you think :)

Currently in staging we're using "ANY"; in production we have multiple webhooks that are all enabled. (See #21 and #415.) 

I think we can merge as-is in safely. 

When we make the swap to using "ANY" webhooks, JBI will be unhealthy; it would be good to communicate this transition, since when we do it we will miss events.

_I believe we had received minimal amount of messages between Saturday night and Sunday afternoon; I could make a swap during this time period to reduce the loss of event data._

leplatrem:
> Currently in staging we're using "ANY"; in production we have multiple webhooks that are all enabled. (See https://github.com/mozilla/jira-bugzilla-integration/issues/21 and https://github.com/mozilla/jira-bugzilla-integration/issues/415.)

That means that when this PR is merged, STAGE's heartbeat will fail right? Or have you deleted all the disabled webhooks entries already?

bsieber-mozilla:
> > Currently in staging we're using "ANY"; in production we have multiple webhooks that are all enabled. (See #21 and #415.)
> 
> That means that when this PR is merged, STAGE's heartbeat will fail right? Or have you deleted all the disabled webhooks entries already?

All deleted!
Screenshot 2023-03-06 at 11 39 38 AM

Issue #58, opened 2022-05-12
Update Contact config as Bugzilla User IDs 


bsieber-mozilla:
Yeah, not sure what it is--was pulling this from the firestore config; we'll have to reach out to Julien to go through the config together and update this file with contact and description info. 

I'll make it an issue to track here.

_Originally posted by @bsieber-mozilla in https://github.com/mozilla/jira-bugzilla-integration/pull/57#discussion_r871541838_

bsieber-mozilla:
- [ ] AMO - https://people.mozilla.org/p/mixedpuppy
- [ ] FIDEDI -  https://people.mozilla.org/p/amir
- [ ] FIDEFE - (+) https://people.mozilla.org/p/yshash
- [ ] flowstate - (+) https://people.mozilla.org/p/pluk
- [ ] fxatps - https://people.mozilla.org/p/pbz
- [ ] fxcm -  (Team split - investigate usage)  https://people.mozilla.org/p/abuzali
- [ ] fxsync - https://people.mozilla.org/p/janet
- [ ] gv - 
- [ ] mv3 - https://people.mozilla.org/p/mixedpuppy
- [ ] nimbus - https://people.mozilla.org/p/janet
- [ ] proton - (Julien suggests to remove tag)
- [ ] relops - 
- [ ] snt - investigate usage / lead engineer

bsieber-mozilla:
Is there any reason we'd not want to put their addresses in the config...?

_I feel like we should inform all these users and request an address for their flows; instead of just using their listed ones._ 

@mozilla/jbi-core thoughts?

bsieber-mozilla:
Is this data planning on being used for automation? (automated email alerting?)

bsieber-mozilla:
Can we migrate configuration to an alternate repo? (@grahamalama had mentioned this pattern previously)

bsieber-mozilla:
> Frida
>  Hello Bryan, how are you? we received a security report related to the integration between Jira and Bugzilla, are the below endpoints intentionally public?
> https://prod.jbi.prod.cloudops.mozgcp.net/whiteboard_tags/
> https://prod.jbi.prod.cloudops.mozgcp.net/jira_projects/
> https://dev.jbi.nonprod.cloudops.mozgcp.net/whiteboard_tags/
> https://dev.jbi.nonprod.cloudops.mozgcp.net/jira_projects/
[...]

PR #440, opened 2023-03-22
Config contact as bugzilla user id

bsieber-mozilla:
To resolve: #58 

Found the user_id's by using the following url: https://bugzilla.mozilla.org/user_profile?login=<email>

...and looking at the Network calls to see the associated user-id.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Below is a full example of an action configuration:
 ```yaml
 - whiteboard_tag: example
   allow_private: false
-  contact: example@allizom.com
+  bugzilla_user_id: 514230
   description: example configuration
   module: jbi.actions.default
   parameters:
@@ -36,10 +36,10 @@ A bit more about the different fields...
     - If false bugs that are not public will not be synchronized. Note that in order to synchronize
       private bugs the bugzilla user that JBI runs as must be in the security groups that are making
       the bug private.
-- `contact`
-    - an email address, a list of email addresses, or a literal "tbd" to signify that no contact is available
-    - If an issue arises with the workflow, communication will be established with these contacts
-    - Please enter the contact information for one or more stakeholders
+- `bugzilla_user_id`
+    - a bugzilla user id, a list of user ids, or a literal "tbd" to signify that no bugzilla user id is available
+    - If an issue arises with the workflow, communication will be established with these users
+    - Please enter the user information for one or more stakeholders
 - `description`
     - string
     - Please enter a description; for example, team name or project use-case.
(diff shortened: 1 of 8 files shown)


config/config.nonprod.yaml
```diff
@@ -1,14 +1,14 @@
 ---
 # Action Config
 - whiteboard_tag: nonprodtest
-  contact: bsieber@mozilla.com
+  contact: 644672
```
leplatrem, line 4: Couldn't we have sub-fields instead? 
Like
```yaml
  contact:
    email: bsieber@mozilla.com
    bugzilla: 644672
```

bsieber-mozilla: We could, but emails are PII and in the linked bug, Frida suggested to modify it to bugzilla user IDs

leplatrem: Ok then.
Yes, renaming the field would definitely help.
`bugzilla_user_id` is a bit ugly compared to `contact` but at least it would be more explicit for readers/contributors. There could be nicer alternative names! 🤷

bsieber-mozilla: Updated from `contact` to `bugzilla_user_id` and updated the readme

README.md
```diff
@@ -256,7 +256,7 @@ GET /whiteboard_tags/
 {
     "addons": {
         "whiteboard_tag": "addons",
-        "contact": "example@allizom.com",
+        "bugzilla_user_id": "514230",
```
leplatrem, line 259: Wouldn't it come out as integer?

jbi/models.py
```diff
@@ -36,8 +28,8 @@ class Action(YamlModel):
 
     whiteboard_tag: str
     module: str = "jbi.actions.default"
-    # TODO: Remove the tbd literal option when all actions have contact info # pylint: disable=fixme
-    contact: EmailStr | list[EmailStr] | Literal["tbd"]
+    # TODO: Remove the tbd literal option when all actions have bugzilla_user_id info # pylint: disable=fixme
+    bugzilla_user_id: int | list[int] | Literal["tbd"]
```
leplatrem, line 32: As far as I can see we didn't test the list of `int`

tests/unit/test_models.py
```diff
@@ -10,7 +10,7 @@ def test_model_serializes():
     action = Action.parse_obj(
         {
             "whiteboard_tag": "devtest",
-            "contact": "person@example.com",
+            "bugzilla_user_id": "123456",
```
leplatrem, line 13: Shouldn't it be integer instead of `str`?

leplatrem (approved):
Just a few fixes for consistency between int and str

README.md
bsieber-mozilla: Fixed!

PR #451, opened 2023-03-28
Using Annotation to reduce duplication of Depends-logic in router.py

bsieber-mozilla:
In new FastAPI release, there was an addition to use `Annotation` for reducing duplicated code; the release notes can be read [here](https://github.com/tiangolo/fastapi/releases)

also check out #443

### DIFF
--- a/jbi/router.py
+++ b/jbi/router.py
@@ -15,14 +15,14 @@ from jbi.models import Actions, BugzillaWebhookRequest
 from jbi.runner import IgnoreInvalidRequestError, execute_action
 from jbi.services import bugzilla, jira
 
-CurrentSettings = Annotated[Settings, Depends(get_settings)]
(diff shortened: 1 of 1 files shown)


grahamalama (approved):
I like it. 

Do we need the `Current` prefix for these annotations? It makes sense in [context of the release notes example](https://github.com/tiangolo/fastapi/releases/tag/0.95.0), but not so much here.

bsieber-mozilla:
> I like it.
> 
> Do we need the `Current` prefix for these annotations? It makes sense in [context of the release notes example](https://github.com/tiangolo/fastapi/releases/tag/0.95.0), but not so much here.

I agree the `Current` isn't great; I've dropped this and left the depends-annotations elements with a "_" prefix

PR #472, opened 2023-04-12
Fix #475, #382: Fix lookup of actions when tag contains dash

### LINKED ISSUE #475: Currently there's a limitation in the lookup code that requires the whiteboard to adhere to a few rules:
Currently there's a limitation in the lookup code that requires the whiteboard to adhere to a few rules:
- `[XYZ-` is what JBI looks for in Bugzilla as the routing key for actions
- XYZ is the name of the whiteboard_tag

`data-platform-infra-wg` and `data-quality` are currently not routable/processable in how JBI [looks up](https://github.com/mozilla/jira-bugzilla-integration/blob/8ea317ac4fb66115e968eab297aa04361d850eaa/jbi/models.py#L305-L311) actions; they'd both map to an action "data" in the current version of jbi.

`dataplatform`: this captures [dataplatform-abc-efg-...] as well as [dataplatform]
`dataquality`: this captures [dataquality-abc-efg-...] as well as [dataquality]

these tags could capture the desired functionality with the given system--albeit they might be removed from current processes.

_Originally posted by @bsieber-mozilla in https://github.com/mozilla/jira-bugzilla-integration/pull/470#discussion_r1163348833_

bsieber-mozilla:
Resolves: #475 

Updates to lookup_action logic-- instead of constructing tags from the whiteboard, then trying to see which tag exists as an action. In this commit we attempt to use the action.whiteboard_tag as the element which must exist within the bug.whiteboard element--iterating through the actions, and acting on the first action that applies to the bug. The whiteboard_tag presented in the action must be contained within the bug’s whiteboard field with a prefix of [

As is: the functionality is working with this new process for action confirmation. 

- [x] Fix logs so that when ActionNotFound is raised and used later for processing, this makes sense in the messaging

This PR ultimately allows for [exact-match-or-partial-matchs-123]:
- `data-eng-platform` and `data-eng-quality` would be able to coexist and map to separate processes


tests/unit/test_runner.py
```diff
@@ -58,6 +58,34 @@ def test_request_is_ignored_because_private(
     assert str(exc_info.value) == "private bugs are not valid for action 'devtest'"
 
 
+def test_request_(
```
bsieber-mozilla, line 61:
```suggestion
def test_request_extended_whiteboard_match(
```

tests/unit/test_router.py
```diff
@@ -125,7 +125,7 @@ def test_webhook_is_200_if_action_raises_IgnoreInvalidRequestError(
         assert response.status_code == 200
         assert (
             response.json()["error"]
-            == "no action matching bug whiteboard tags: unmatched"
+            == "no action matching bug whiteboard tags: devtest"
```
bsieber-mozilla, line 128: this log makes a little less sense with this transition

bsieber-mozilla:
```suggestion
            == "no whiteboard matching action tags: devtest"
```

leplatrem:
Bryan, could you explain the reason behind this change? 
It may be a good opportunity to start a `docs/adrs/` folder, and add a brief one for this. 

> Side nitpicking: there's no issue associated to this ;)

bsieber-mozilla:
> Bryan, could you explain the reason behind this change?
> It may be a good opportunity to start a docs/adrs/ folder, and add a brief one for this.

We have an action that currently is impossible to map whiteboard tags to with the current `.split` process that occurs on the `-` character: 
https://github.com/mozilla/jira-bugzilla-integration/blob/846d69a6d661065ac2fcfe802f8548ac34123d6b/config/config.prod.yaml#L221

Currently any whiteboard tag that comes in with: `[remote-settings]` gets assigned the action_tag `remote` as the lookup key in the action dictionary.

Reversing the lookup to find the action_tag inside the whiteboard_tag without parsing the whiteboard_tag as heavily; enables this action as well as future actions to have "-"s in them. 

In the later commits, I added `brackets_required` to the Action model; the default behavior of having the brackets required can now be disabled per action. This has been implicit and is now likely a process many users follow.

This PR now makes [this behavior] something we are not requiring, but are defaulting to at this time.

bsieber-mozilla:
>     Side nitpicking: there's no issue associated to this ;)

Ahh, upon second reading this made sense; I thought you were mentioning there's no issue with the code.. but there's no github issue associated with this change, is the conclusion I just came to... I made #475 from the DENG comment, forcing a user to adjust their whiteboard processing. 

I'll add the above comment in regards to [remote-settings] to the gh issue as well

tests/conftest.py
```diff
@@ -225,11 +225,35 @@ def action_example() -> Action:
...
+    return Actions.parse_obj([action_example_with_inner_match])
+
+
+@pytest.fixture
+def actions_example_without_brackets_required(
+    action_example_without_brackets_required,
+) -> Actions:
+    return Actions.parse_obj([action_example_without_brackets_required])
+
+
```
grahamalama, line 256: If these fixtures are only used in one or two tests, it may be better to define them at the module level, rather than clutter the root `conftest.py`.

Similarly, it looks like `action_example_with_inner_match` is only used to build `actions_example_with_inner_match`. Maybe we can just do

```python
@pytest.fixture
def actions_example_with_inner_match() -> Actions:
    action = action_factory(whiteboard_tag="inner-match")
    return Actions.parse_obj([action])
```

repeating that pattern for `action_example_without_brackets_required` / `actions_example_without_brackets_required`.

grahamalama: I'm also noticing that these factories would be more useful if we had fixtures that simply returned the factory function, e.g. 

```
@pytest.fixture
def action_factory() -> Action:
    return factories.action_factory
```

Then we could use them directly in tests.

tests/unit/test_runner.py
grahamalama, line 1: Should we be testing the changes of this PR somewhere in `tests/unit/test_bugzilla.py`, maybe similar to [these](https://github.com/mozilla/jira-bugzilla-integration/blob/044e4a21a1d8259b2c99ba2be13aa7844a4a1b64/tests/unit/test_bugzilla.py#L56-L67) tests? Testing at the runner level perhaps feels too "wide" to make assertions about when it seems like we're asserting specific behavior of how `BugzillaBug.lookup_action` works.

tests/conftest.py
bsieber-mozilla:
```
@pytest.fixture
def action_factory() -> Action:
    return factories.action_factory
```

Initially there was some overloading issues when trying this out; but I modified the import practices to allow for this naming convention (and usage in the tests).

As for the repeating of tests, the newest commit is aimed to address that concern as well.

tests/unit/test_runner.py
bsieber-mozilla: I added some more tests to `test_bugzilla`; I agree `test_runner` might be a bit "wide" but I do believe the new tests coexist with some of the prior testing strategy/process in the `test_runner`

tests/conftest.py
```diff
@@ -206,54 +206,37 @@ def webhook_change_status_assignee():
...
 @pytest.fixture
-def action_example_with_inner_match() -> Action:
-    return action_factory(whiteboard_tag="inner-match")
-
-
-@pytest.fixture
-def action_example_without_brackets_required() -> Action:
-    return action_factory(whiteboard_tag="bracket-less", brackets_required=False)
+def action_example() -> Action:
+    return factories.action_factory()
```
bsieber-mozilla: _with the usage of `action_factory` from the `factories` module being imported to the namespace; the name `action_factory` was not possible in this file prior._

ref: https://github.com/mozilla/jira-bugzilla-integration/pull/472#discussion_r1167237261

README.md
```diff
@@ -34,9 +34,15 @@ A bit more about the different fields...
...
     - bool [true, false]
     - default: false
-    - If false bugs that are not public will not be synchronized. Note that in order to synchronize
+    - If `false` bugs that are not public will not be synchronized. Note that in order to synchronize
       private bugs the bugzilla user that JBI runs as must be in the security groups that are making
       the bug private.
+- `brackets_required` (optional)
+    - bool [true, false]
+    - default: true
+    - If `true` the whiteboard_tag must be be contained within an opening ([) and closing (]) bracket, this is the default behavior; i.e. this tag `XYX` matches this whiteboard_tag `[XYZ]` but not `XYZ`.
```
leplatrem, line 43: Something that I figured out reading the code as well: it would match: `[xyz]`, `[something-XYZ]`, `[XYZ-something]`

tests/unit/test_bugzilla.py
```diff
@@ -60,11 +61,43 @@ def test_lookup_action(actions_example):
...
+    action = bug.lookup_action(actions_example)
+    assert action.whiteboard_tag == "devtest"
+    assert "test config" in action.description
+
+
+def test_lookup_action_with_prefix_and_postfix(actions_example):
+    bug = bug_factory(id=1234, whiteboard="[example][foo-DevTest-bar]")
+    action = bug.lookup_action(actions_example)
+    assert action.whiteboard_tag == "devtest"
+    assert "test config" in action.description
```
leplatrem, line 82: Since these 3 tests are constructed the same way, you can use `@pytest.parameterize` passing the whiteboard.

jbi/models.py
```diff
@@ -304,11 +306,15 @@ def extract_from_see_also(self):
...
-        tags: list[str] = self.get_potential_whiteboard_config_list()
-        for tag in tags:
-            if action := actions.get(tag):
-                return action
-        raise ActionNotFoundError(", ".join(tags))
+        if self.whiteboard:
+            for tag, action in actions.by_tag.items():
+                pre = r"\[.*" if action.brackets_required else ""
+                post = r".*\]" if action.brackets_required else ""
+                search_string = f"{pre}{tag.lower()}{post}"
```
leplatrem, line 313: `[.*{tag}.*]` may be too permissive:

```
>>> import re
>>> re.search(r"\[.*devtest.*\]", "[foo] devtest [bar]")
<re.Match object; span=(0, 19), match='[foo] devtest [bar]'>
>>>
```

you would have to change it to something like this:

```
>>> re.search(r"\[[^\]]*devtest[^\]]*\]", "[foo] devtest [bar]")
>>>
>>> re.search(r"\[[^\]]*devtest[^\]]*\]", "[foo-devtest-bar]")
<re.Match object; span=(0, 17), match='[foo-devtest-bar]'>
>>>
```
≈

README.md
```diff
@@ -34,9 +34,15 @@ A bit more about the different fields...
 - `allow_private` (optional)
     - bool [true, false]
     - default: false
-    - If false bugs that are not public will not be synchronized. Note that in order to synchronize
+    - If `false` bugs that are not public will not be synchronized. Note that in order to synchronize
       private bugs the bugzilla user that JBI runs as must be in the security groups that are making
       the bug private.
+- `brackets_required` (optional)
```
leplatrem, line 40: If this option really necessary? I ask because no configuration entry was modified in this PR

jbi/models.py
```diff
@@ -304,11 +306,15 @@ def extract_from_see_also(self):
...
+                if re.search(search_string, self.whiteboard.lower()):
```
leplatrem, line 314: nitpick: instead of turning both inputs to lowercase, you can do: `if re.search(search_string, self.whiteboard, flags=re.IGNORE_CASE):`

grahamalama (approved):

PR #494, opened 2023-05-03
[deng] reduce customization of jbi workflow

bsieber-mozilla:


bsieber-mozilla:
Forcing this configuration change through

Issue #345, opened 2022-12-15
Unhandled `ConnectionError` in `/__heartbeat__`


grahamalama:
[This](https://sentry.io/organizations/mozilla/issues/3750108748/?project=6364263) seems to be the most common error in Sentry at the moment:
```
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

It's coming from the `/__heartbeat__` route and appears to be related to the `bugzilla` service. That's the only service that appears in the stacktrace, anyway.

If we time out when trying to reach Bugzilla, should we return a `false` for the healthcheck?

PR #503, opened 2023-05-15
Log HTTP errors responses of Jira client

leplatrem:


jbi/services/jira.py
```diff
@@ -30,15 +31,45 @@
 instrumented_method = instrument(prefix="jira", exceptions=(errors.ApiError,))
```
grahamalama, line 31: This is a drive by fix, but I noticed when investigating the `log_http_error` decorator that the Jira client doesn't ever raise `errors.ApiError` -- it only raises `request.HTTPError`s. Should we change this line to reflect that?

jbi/services/jira.py
```diff
@@ -30,15 +31,45 @@
...
-    set_issue_status = instrumented_method(Jira.set_issue_status)
-    issue_add_comment = instrumented_method(Jira.issue_add_comment)
-    create_issue = instrumented_method(Jira.create_issue)
+    update_issue_field = log_http_error(instrumented_method(Jira.update_issue_field))
+    set_issue_status = log_http_error(instrumented_method(Jira.set_issue_status))
+    issue_add_comment = log_http_error(instrumented_method(Jira.issue_add_comment))
+    create_issue = log_http_error(instrumented_method(Jira.create_issue))
+    get_project_components = log_http_error(
+        instrumented_method(Jira.get_project_components)
+    )
```
grahamalama, line 72: Could we get away with only monkeypatching the base class's [`raise_for_status`](https://github.com/atlassian-api/atlassian-python-api/blob/32b5af22ad3702a046b6184506fc1b5bd98a7b40/atlassian/rest_client.py#L462) method?

```suggestion
    raise_for_status = log_http_error(Jira.raise_for_status)
    update_issue_field = instrumented_method(Jira.update_issue_field)
    set_issue_status = instrumented_method(Jira.set_issue_status)
    issue_add_comment = instrumented_method(Jira.issue_add_comment)
    create_issue = instrumented_method(Jira.create_issue)
    get_project_components = instrumented_method(Jira.get_project_components)
```

Alternatively, we could write our own `raise_for_status` since we'd potentially double log the error in [this case](https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/rest_client.py#LL479C1-L481C44):

```
except Exception as e:
    log.error(e)
    response.raise_for_status()
```

jbi/services/jira.py
leplatrem: OH, but that's probably why we have #345 !

jbi/services/jira.py
leplatrem: Yes, you're right, let's decorate `raise_for_status()`, it's way more elegant!

> since we'd potentially double log the error in [this case](https://github.com/atlassian-api/atlassian-python-api/blob/master/atlassian/rest_client.py#LL479C1-L481C44):

Yes, this `Exception` is raised when the response cannot be parsed into JSON or when the parsed content does not have the right format. If I'm not wrong, we would indeed have two calls to `logger.error()`, but not with the same error. One with the JSON/parsing stuff, and one with the HTTP status.

jbi/services/jira.py
grahamalama: Nice! Good catch.

grahamalama (approved):

grahamalama (approved):

Issue #517, opened 2023-05-18
Decouple Jira issue creation and updates from whiteboard tag to issue label syncing


grahamalama:
JBI currently syncs the whiteboard tags of a bug with Jira issue labels when they are created or updated. This assumes that:
- the Jira automation service account has the ability to add labels to Jira issues
- the project is configured to include labels as a field on the Create / Update issue [screens](https://support.atlassian.com/jira-cloud-administration/docs/configure-issue-screens/#Configure-a-screen--x27-s-tabs-and-fields)

Failing to account for these assumptions being false leads to issues with creating issues, as evidenced in this [Sentry alert](https://mozilla.sentry.io/issues/3571741650/?project=6364263)

To prevent potential syncing failures, we should modify the current workflow. This ticket aims to decouple the syncing of whiteboard tags and issue labels from creating or updating issues. By doing so, even if the assumptions mentioned above are not met, the creation of an issue will not be blocked.

grahamalama:
@leplatrem thinking about this a bit more, I'm wondering if we're leading ourselves down a path where we send double the amount of requests we need. Right now, here's the flow of creating an issue when `sync_whiteboard_labels` is enabled:
<details><summary>Current Issue creation flow</summary>

```mermaid
sequenceDiagram
    Bugzilla ->> JBI: POST /bugzilla_webhook (bug.create payload)
    JBI ->> Jira: POST /issue (Create issue, which includes labels)
    Jira -->> JBI: HTTP 2xx (Issue created)
    JBI -->> Bugzilla: HTTP 2xx (Webhook processed)
```

</details> 

As we discovered, this isn't very resilient. We need to check that we're _able_ to add labels first. With the `maybe_update_labels` approach, I think that'd look like:
<details><summary>Create issue, then sync labels</summary>

```mermaid
sequenceDiagram
    Bugzilla ->> JBI: POST /bugzilla_webhook (bug.create payload)
    JBI ->> Jira: POST /issue (Create Issue)
    Jira -->> JBI: HTTP 2xx (Issue created)
    JBI ->> Jira: GET /createmeta (Check ability to add labels)
    Jira -->> JBI: HTTP 2xx (Ability to add labels confirmed)
    JBI ->> Jira: PUT /issue/<issue_id>/labels (Add whiteboard tags as labels)
    Jira -->> JBI: HTTP 2xx (Labels added)
    JBI -->> Bugzilla: HTTP 2xx (Webhook processed)
```

</details> 

[...]

PR #520, opened 2023-05-19
Fix #518: add `maybe_update_components`

### LINKED ISSUE #518: Create a `maybe_sync_component` step to pull BZ component into Jira components
User story: as an engineering manager, I sometimes want to be able to filter and organize Jira tickets using Jira's `components` field. It would be helpful to have that `components` field populated with the corresponding component from the Bugzilla ticket. (probably in the common format of `Product::Component`).

Example: [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1825312) is synced to the FXDROID project using the `[fxdroid]` whiteboard, resulting in [this ticket](https://mozilla-hub.atlassian.net/browse/FXDROID-363). The Jira ticket's `components` field is empty.

Notes:
* One challenge here is that - AFAIK - "components" in Jira has to be a predefined set of items. So components would need to be created in both places ([Jira project admin](https://mozilla-hub.atlassian.net/plugins/servlet/project-config/FXDROID/administer-components) and Bugzilla admin) before being able to be synced.
* As a temporary workaround, my team can just copy the component into a whiteboard, which will sync it to the Jira ticket as a 'label'. That at least allows us to do much of the filtering we need to do (it just requires that manual step of adding the correct whiteboard)

leplatrem:
I found #518 interesting to work on, since it is related to what we are trying to achieve in #517.

I decoupled the setting of the `components` field from the `create_issue` step, and made it a dedicated "maybe step". 
If the bug component isn't part of the existing Jira project components, a warning is issued and the step does not crash (hence the `maybe_`)

I believe that this approach (removing parameters from `create_issue`) follows more what we described in our first ADR https://github.com/mozilla/jira-bugzilla-integration/blob/main/docs/adrs/001-actions-reusability.md

### DIFF
--- a/jbi/actions/steps.py
+++ b/jbi/actions/steps.py
@@ -203,7 +203,9 @@ def maybe_update_components(context: ActionContext, **parameters):
     Update the Jira issue components
     """
     config_components: list[str] = parameters.get("jira_components", [])
-    candidate_components = set([context.bug.component] + config_components)
+    candidate_components = set(config_components)
+    if context.bug.component:
+        candidate_components.add(context.bug.component)
     client = jira.get_client()
 
     # Fetch all projects components, and match their id by name.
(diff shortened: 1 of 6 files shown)


jbi/actions/steps.py
```diff
@@ -198,3 +196,37 @@ def maybe_update_issue_status(context: ActionContext, **parameters):
             return context, (resp,)
 
     return context, ()
+
+
+def maybe_update_components(context: ActionContext, **parameters):
+    """
+    Update the Jira issue components
+    """
+    config_components: list[str] = parameters.get("jira_components", [])
```
grahamalama, line 205: I wonder if there's a better way to handle this than casting. What if someone makes `jira_components` a comma-delimited string, for example. Hopefully something we'd catch in code review for config changes, but something could slip through.

We could maybe type the parameters object or include a guard to check that `jira_components` is a list, emitting a warning if it isn't.

leplatrem: That's already how it's done on `main`. I opened https://github.com/mozilla/jira-bugzilla-integration/pull/523 for that

Issue #393, opened 2023-02-08
Updating a bug removes any labels added directly in Jira


davehunt:
I noticed that any labels added in Jira before linking it to a bug were removed when the initial sync happened. Ideally, the sync would merge the existing labels with those from the whiteboard. I realise this would mean that removing whiteboard entries will leave them as labels in Jira, but this is less destructive than the current behaviour.

bsieber-mozilla:
related: https://github.com/mozilla/jira-bugzilla-integration/issues/378

bsieber-mozilla:
@mozilla/jbi-core; looking for some thoughts/feedback here. 

I'm thinking we move forward with option 1 presented below; that way the source of truth is determined by the flag--but initial reflection allows for usage of labels/tags from bz for triage

----

https://github.com/mozilla/jira-bugzilla-integration/issues/378#issue-1557166920


> There’s likely other options; but there a few that come to mind:
> 1. Always have whiteboard_labels constructed on creation events; only use flag on "modify" events
> 2. Separate `sync_whiteboard_labels` flag, appending  `_on_create` and `_on_update`
>     a. Issues can be constructed with whiteboard labels--but not overwritten in jira if `_on_update`  is false
> 3. Change name to `overwrite_jira_labels`
>     a. if current behavior is maintained
> 4. Change behavior to pull latest Jira labels
>     a. requires additional API call
> ```
>         jira_labels = context.jira.issue_field_value(key=issue_key, field="labels")
> ```

PR #527, opened 2023-05-22
Fix #517: decouple labels from `create_issue` and `update_issue` steps

### LINKED ISSUE #517: Decouple Jira issue creation and updates from whiteboard tag to issue label syncing

leplatrem:
This PR does the strict minimum:
- move the labels setting out of `create_issue` and `update_issue`
- catch and log 400 errors, without crashing, to prevent to block the queue

It does not:
- try to be smart about create/issue meta fields
- try to limit the number of calls to the Jira API
- merge the existing labels with the new ones (#393


grahamalama (commented):
Nice! I'm glad we're moving in this direction to make everything more modular. 

At this point, all `update_issue` and `update_jira_issue` do is update an issue's summary. Should we be explicit about and rename those to `update_summary` and `update_jira_issue_summary` accordingly?

grahamalama (approved):

PR #541, opened 2023-05-25
Fix #531: count partial executions of workflows

### LINKED ISSUE #531: Add instrumentation for partial syncs
Creating or updating a Jira issue based on changes to a bug can involve one or more steps, with each step usually referring to one aspect of the Jira issue to change based on the aspect of a bug (e.g. `maybe_update_resolution` or `maybe_update_status`).

We've designed action workflows so that if one step fails, it does not prevent the rest of the workflow from running.

For this ticket, we should somehow track how many "partial" syncs we perform where not all of the steps for a workflow succeed.

leplatrem:


jbi/actions/default.py
```diff
@@ -97,7 +99,15 @@ def __call__(self, context: ActionContext) -> ActionResult:
...
         for step in self.steps[context.operation]:
-            context, step_responses = step(context=context, **self.parameters)
+            try:
+                context, step_responses = step(context=context, **self.parameters)
+            except Exception:
+                if len(responses) > 0:
+                    # Count the number of workflows that produced at least request,
+                    # but could not complete entirely successfully.
+                    statsd.incr("jbi.bugzilla.partial.count")
+                raise
```
grahamalama, line 109: I think this doesn't exactly cover what I was envisioning, though it could still be valuable to capture 🤔. 

For what I was thinking, say a user specifies these steps:
```python
"new": [
    "create_issue",
    "maybe_update_components",
],
```

The issue is created, but when we go to add the component that the user specified in `jira_components`, it isn't on the project for some reason, which hits this branch:

```python
if candidate_components:
    logger.warning(
        "Could not find components %s in project",
        candidate_components,
        extra=context.dict(),
    )
```

In this case, all of the steps in the workflow have run and no exceptions were raised, but the end result of the workflow is an issue that is missing some data the user expected to be there. The user might ask "why wasn't the component set? I have that in the config."

In this case, I'd consider the sync "incomplete", and increase `jbi.bugzilla.partial.count`. It's kind of a signal that while we didn't hit any exceptions while we were trying to complete the sync, we lost some data due to errors or misconfiguration along the way

jbi/actions/default.py
```diff
@@ -96,8 +98,23 @@ def __call__(self, context: ActionContext) -> ActionResult:
...
+
         for step in self.steps[context.operation]:
-            context, step_responses = step(context=context, **self.parameters)
+            try:
+                context, step_responses = step(context=context, **self.parameters)
+            except Exception:
+                if len(responses) > 0:
+                    # Count the number of workflows that produced at least request,
+                    # but could not complete entirely successfully.
+                    statsd.incr("jbi.bugzilla.aborted.count")
```
grahamalama, line 110: Should we add the name of the workflow somewhere in the name of this stat? Might be helpful to see which workflows are aborted or finish as incomplete the most.

jbi/actions/default.py
```diff
@@ -96,8 +98,23 @@ def __call__(self, context: ActionContext) -> ActionResult:
...
+                raise
+
+            if not step_responses:
+                # If the step did not generate HTTP response, it means the step
+                # skipped its operation (for various reasons, safety checks, etc.)
+                incomplete = True
```
grahamalama, line 116: Not generating an HTTP response might not always be a signal that the step didn't complete as expected. Take:

```python
def update_issue_summary(context: ActionContext, **parameters):
    """Update the Jira issue's summary if the linked bug is modified."""

    bug = context.bug
    issue_key = context.jira.issue

    if "summary" not in context.event.changed_fields():
        return context, ()
```

Us returning a context object early with no response isn't a signal that something went wrong -- it just means the summary wasn't updated, so we don't need to do anything to the linked issue. In this case, the step function behaved 100% correctly.

jbi/actions/default.py
```diff
@@ -95,15 +97,30 @@ def __init__(self, steps: dict[Operation, list[Callable]], **parameters):
...
+
         for step in self.steps[context.operation]:
             context = context.update(current_step=step.__name__)
             try:
                 context = step(context=context, **self.parameters)
             except IncompleteStepError as exc:
                 # Step did not execute all its operations.
                 context = exc.context
+                statsd.incr(
+                    f"jbi.action.{context.jira.project.lower()}.incomplete.count"
```
grahamalama, line 110: Can we use the whiteboard tag of the workflow instead of the jira project, since the whiteboard tag is more what we consider the "name" of the workflow? Same below in the `aborted` case

jbi/actions/default.py
```diff
@@ -95,15 +97,30 @@ def __init__(self, steps: dict[Operation, list[Callable]], **parameters):
...
+                    # Count the number of workflows that produced at least one request,
+                    # but could not complete entirely with success.
+                    statsd.incr(
+                        f"jbi.action.{context.jira.project.lower()}.aborted.count"
+                    )
+                raise
 
             step_responses = context.responses_by_step[step.__name__]
+            if step_responses:
+                has_produced_request = True
```
grahamalama, line 123: I'm having a hard time seeing how this `has_produced_request` variable works to determine if a workflow was aborted or not. What does this signal vs just incrementing when we hit an `Exception`?

tests/unit/actions/test_default.py
```diff
@@ -95,3 +96,47 @@ def test_default_returns_callable_with_data(
...
+            callable_object(context=context_create_example)
+
+    mocked.incr.assert_called_with("jbi.action.fenix.aborted.count")
+
+
+def test_counter_is_incremented_when_workflows_was_incomplete(
+    context_create_example: ActionContext, mocked_bugzilla, mocked_jira
+):
+    context_create_example.jira.project = "FENIX"
+    context_create_example.bug.resolution = "WONTFIX"
```
grahamalama, line 122: This is more a matter of taste, but in these cases where we're overwriting properties of some "example" object, it seems like we could also use a factory.

jbi/actions/default.py
leplatrem: I was thinking of this:
- if a workflow crashes before or during its first request, nothing was executed on the Jira side
- if a workflow crashes after a request was executed, the workflow was partially executed, "aborted", and the state on the Jira side is potentially inconsistent
- if we would count exceptions it's more or less equivalent to counting 5XX responses

grahamalama: Ahh okay, thanks for that clarification

PR #543, opened 2023-05-26
Fix errors observed after v4.4.0 release

grahamalama:
In https://github.com/mozilla/jira-bugzilla-integration/commit/c178e304b0858bba15c6a60c15ddbd46fb7cf327, we added typing for action parameters, which made it more "ergonomic" to access params and know their types.

However, this lead to [some](https://mozilla.sentry.io/issues/4208127241/?project=6364263&query=is%3Aunresolved&referrer=issue-stream&stream_index=1) [issues](https://mozilla.sentry.io/issues/4208116624/?project=6364263&query=is%3Aunresolved&referrer=issue-stream&stream_index=0)  in some functions where we'd `get()` params, providing a default if the key wasn't in the params dict.

Here is a subset of `ActionParameters`:

```python
    status_map: Optional[dict[str, str]] = None
    resolution_map: Optional[dict[str, str]] = None
    jira_components: Optional[list[str]] = None
```

And here's a snippet where one of the errors popped up:
```python
resolution_map: dict[str, str] = parameters.get("resolution_map", {})
jira_resolution = resolution_map.get(context.bug.resolution or "")
```
In this case, we expect to set `resolution_map` to `{}` if it isn't set in the config. But based on how the Pydantic model behaves, if `resolution_map` is unset in the config, it gets set to `resolution_map=None` in the Pydantic model, so we set `resolution_map` to `None`.

Here's where the other error popped up:

```python
    components_by_project = {
        action.parameters.jira_project_key: action.parameters.jira_components
        for action in actions
    }
    success = True
    for project, specified_components in components_by_project.items():
        all_project_components = get_client().get_project_components(project)
        all_components_names = set(comp["name"] for comp in all_project_components)
        unknown = set(specified_components) - all_components_names
```

In this case, we still need a guard to filter out actions where `parameters.jira_components` is `None`, because we can't make a `set` out of None.

For now, my strategy was to revert the PR where we added this `ActionParameters` object and add some regression test cases.

While I was here, I took the opportunity to fix some other issues I saw.

Also, now that I'm looking again at #523, `ActionParameters` might be going in a direction that's different than was intended as far as how "free" users are to supply entirely custom actions and parameters. I'll open this discussion in a new issue, but it might warrant an ADR if we want to narrow action parameters to only certain fields / types.

### DIFF
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -7,7 +7,7 @@ is created or updated, its `operation` attribute will be `Operation.CREATE` or `
 and when a comment is posted, it will be set to `Operation.COMMENT`.
 """
 import logging
-from typing import Optional
+from typing import Callable, Optional
 
 from jbi import ActionResult, Operation
 from jbi.actions import steps as steps_module
@@ -86,7 +86,7 @@ def init(
 class Executor:
     """Callable class that encapsulates the default action."""
 
-    def __init__(self, steps, **parameters):
+    def __init__(self, steps: dict[Operation, list[Callable]], **parameters):
         """Initialize Executor Object"""
         self.steps = steps
         self.parameters = parameters
--- a/jbi/actions/steps.py
+++ b/jbi/actions/steps.py
@@ -5,7 +5,7 @@ Each step takes an `ActionContext` and a list of arbitrary parameters.
 """
 
 import logging
-from typing import Optional
+from typing import Iterable, Optional
 
 from requests import exceptions as requests_exceptions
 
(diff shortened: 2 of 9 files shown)


leplatrem (approved):
> Now that I'm looking again at https://github.com/mozilla/jira-bugzilla-integration/pull/523, ActionParameters might be going in a direction that's different than was intended as far as how "free" users are to supply entirely custom actions and parameters.

I agree, let's discuss it in #544 
There's more to that, it's not very elegant if typings of steps parameters are spread into `models.py`.

PR #555, opened 2023-06-07
Fix #552: do not match tags if not surrounded by dashes

### LINKED ISSUE #552: Whiteboard tags match is too lax
`[domcore]` matches `omc` and shouldn't

Related issue #472

leplatrem:


grahamalama (approved):

tests/unit/test_bugzilla.py
```diff
@@ -31,10 +31,16 @@ def test_extract_see_also(see_also, expected):
...
+        "[test-DevTest]",
+        "[foo-DevTest-bar]",
+        "[foo-bar-DevTest-foo-bar]",
         "[example][DevTest]",
-        "[example][DevTest-test]",
-        "[example][test-DevTest]",
-        "[example][foo-DevTest-bar]",
+        "[DevTest][example]",
+        "[example][DevTest][example]",
     ],
```
grahamalama, line 44: Should we also add test cases for whiteboard strings that include space delimiters as well?

PR #557, opened 2023-06-07
Fix #554: configurable issue type

### LINKED ISSUE #554: Check that all projects have `Bug` and `Task` issue types
To avoid issues like https://mozilla.sentry.io/issues/4231342450/?referrer=slack

Or offer the ability to configure bug/issue type mapping from configuration...

leplatrem:


jbi/actions/steps.py
```diff
@@ -35,12 +35,20 @@ def create_issue(context: ActionContext, **parameters):
...
+        raise TypeError("The 'issue_type_map' parameter must be a dictionary.")
+    issue_type = issue_type_map.get(bug.type, "Task")
+
     # In the payload of a bug creation, the `comment` field is `null`.
     # We fetch the list of comments to use the first one as the Jira issue description.
     comment_list = bugzilla.get_client().get_comments(bug.id)
     description = comment_list[0].text if comment_list else ""
 
-    jira_create_response = jira.create_jira_issue(context, description)
+    jira_create_response = jira.create_jira_issue(context, description, issue_type)
```
grahamalama, line 51: Maybe something to track in a follow-on, but should we have a "pre-flight" check here to ensure that either the configured issue or default that we pick is available as an issue type on the project? This PR allows configuration, but it's possible that even the custom issue type mapping could drift from what's available in Jira.

grahamalama (approved):
Preemptively approving this for when you get that test to pass

jbi/actions/steps.py
leplatrem: https://github.com/mozilla/jira-bugzilla-integration/issues/575

PR #558, opened 2023-06-07
Fix #553: don't retry 4XX errors in Jira client

### LINKED ISSUE #553: Do not retry bad requests HTTP 400
On 400 error responses, the client should not repeat this request without modification.

The decorator we introduced in #152 retries them.

leplatrem:


jbi/services/jira.py
```diff
@@ -30,12 +30,23 @@
 
 JIRA_DESCRIPTION_CHAR_LIMIT = 32767
 
+
+def fatal_code(exc):
+    """Do not retry 4XX errors, mark them as fatal."""
+    try:
+        return 400 <= exc.response.status_code < 500
```
grahamalama, line 37: Do we want to be more granular than this? For example, we'd want to retry on a `429 Too Many Requests` right? That's why we're using the `backoff.expo` strategy, after all.

grahamalama, line 37: Do we want to be more granular than this? For example, we'd want to retry on a `429 Too Many Requests` right? That's why we're using the `backoff.expo` strategy, after all.

leplatrem: > For example, we'd want to retry on a 429 Too Many Requests right? 

Well, that one precisely I don't think so, since we're already sending too many requests :/ 

According to the spec, 4XX shouldn't be retried without changing the request

> That's why we're using the backoff.expo strategy, after all

Partly. If the server returns 503, we don't want to DDOS it more and wait a bit before retrying for example

grahamalama: Ahh, [this](https://github.com/litl/backoff/blob/d82b23c42d7a7e2402903e71e7a7f03014a00076/README.rst#backoffruntime) is interesting

PR #559, opened 2023-06-07
Fix #521: do not add brackets on issue labels by default

### LINKED ISSUE #521: Add option to add labels with or without brackets
> Noting the labels, is it expected that this have both `"fxsync-"` and `"[fxsync-]"?`

Adding an option `labels_brackets: [no|yes|both]` could make sense

leplatrem:


grahamalama (approved):
Can we add one test case where we raise a `ValueError`?

PR #577, opened 2023-06-12
Ref #541: introduce `IncompleteStepError` to distinguish step early exit from handled errors

leplatrem:
Ref #541

### DIFF
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -9,7 +9,8 @@ and when a comment is posted, it will be set to `Operation.COMMENT`.
 import logging
 from typing import Callable, Optional
 
-from jbi import ActionResult, IncompleteStepError, Operation
+from jbi import ActionResult, Operation
+from jbi.errors import IncompleteStepError
 from jbi.actions import steps as steps_module
 from jbi.environment import get_settings
 from jbi.models import ActionContext
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -10,9 +10,9 @@ import logging
 from typing import Callable, Optional
 
 from jbi import ActionResult, Operation
-from jbi.errors import IncompleteStepError
 from jbi.actions import steps as steps_module
 from jbi.environment import get_settings
+from jbi.errors import IncompleteStepError
 from jbi.models import ActionContext
 
 settings = get_settings()
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -9,7 +9,7 @@ and when a comment is posted, it will be set to `Operation.COMMENT`.
 import logging
 from typing import Callable, Optional
 
-from jbi import ActionResult, Operation
+from jbi import ActionResult, IncompleteStepError, Operation
 from jbi.actions import steps as steps_module
 from jbi.environment import get_settings
 from jbi.models import ActionContext
@@ -94,7 +94,11 @@ class Executor:
     def __call__(self, context: ActionContext) -> ActionResult:
         """Called from `runner` when the action is used."""
         for step in self.steps[context.operation]:
-            context = step(context=context, **self.parameters)
+            try:
+                context = step(context=context, **self.parameters)
+            except IncompleteStepError as exc:
+                context = exc.context
+
         for response in context.responses:
             logger.debug(
                 "Received %s",
--- a/jbi/actions/default.py
+++ b/jbi/actions/default.py
@@ -6,6 +6,7 @@ The `runner` will call this action with an initialized context. When a Bugzilla 
 is created or updated, its `operation` attribute will be `Operation.CREATE` or `Operation.UPDATE`,
 and when a comment is posted, it will be set to `Operation.COMMENT`.
 """
+import itertools
 import logging
 from typing import Callable, Optional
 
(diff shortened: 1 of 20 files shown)


grahamalama (commented):
Initial thoughts on a first look:

- Can we add some tests for this new behavior?

- Are there any other functions we could add the `IncompleteStepError` to? I'm thinking most if not all of the  of the `maybe_` steps would be good candidates

- This pattern of
```python
context.update(responses=context.responses + [<response>])
```

might deserve a method for handling this, like `context.append_response` or something

- Would it be useful to include information about what response maps to which step in the responses object? So instead of `responses` just being a list of responses, it might be a dict like `{<step_name>: <step_response>, ...}` or maybe a list of tuples `[(step_name, response), ...]`?

leplatrem:
I'm not entirely sure about the last commit design...

jbi/models.py
```diff
@@ -359,7 +360,25 @@ class ActionContext(Context, extra=Extra.forbid):
 
     rid: str
     operation: Operation
+    current_step: Optional[str]
     event: BugzillaWebhookEvent
     jira: JiraContext
     bug: BugzillaBug
     extra: dict[str, str] = {}
+    responses_by_step: dict[str, list] = {}
```
grahamalama, line 368: I think this will work too?
```suggestion
    responses_by_step: DefaultDict[str, list] = defaultdict(list)
```

jbi/models.py
```diff
@@ -359,7 +360,25 @@ class ActionContext(Context, extra=Extra.forbid):
...
+
+    @property
+    def responses(self):
```
grahamalama, line 371: Is there a reason to have both the `responses_by_step` property and this flattened `responses` representation? Is it was because we didn't want to make the scope of this change too large?

jbi/models.py
```diff
@@ -359,7 +360,25 @@ class ActionContext(Context, extra=Extra.forbid):
...
+        """Return flat list of all responses"""
+        # return list(itertools.chain.from_iterable(self.responses_by_step.values()))
```
grahamalama, line 373: Is this an old implementation and you went for something less complex, i.e. the uncommented code below?

jbi/actions/default.py
```diff
@@ -93,20 +94,20 @@ def __init__(self, steps: dict[Operation, list[Callable]], **parameters):
...
-                    extra={
-                        "response": response,
-                        **context.dict(),
-                    },
-                )
-            responses += step_responses
-
-        return True, {"responses": responses}
+            try:
+                context = context.update(current_step=step.__name__)
```
grahamalama, line 99: Should we move this above the `try` block?

jbi/actions/default.py
```diff
@@ -93,20 +94,20 @@ def __init__(self, steps: dict[Operation, list[Callable]], **parameters):
...
+                context = step(context=context, **self.parameters)
+            except IncompleteStepError as exc:
+                context = exc.context
+
+        for response in context.responses:
```
grahamalama, line 104: I think we'll still want to log after we complete the step and get the responses since in this case, we'd log the same ("complete") context object for each response instead of logging the context as we go.

tests/unit/actions/test_default.py
```diff
@@ -88,7 +88,10 @@ def test_default_returns_callable_with_data(
     mocked_jira.create_or_update_issue_remote_links.return_value = sentinel
     mocked_bugzilla.get_bug.return_value = context_create_example.bug
     mocked_bugzilla.get_comments.return_value = []
-    callable_object = default.init(jira_project_key=context_create_example.jira.project)
+    callable_object = default.init(
+        jira_project_key=context_create_example.jira.project,
+        steps={"new": ["create_issue", "add_link_to_bugzilla"]},
```
grahamalama, line 93: I'm curious, what was the motivation for this change?

tests/unit/actions/test_steps.py
```diff
@@ -305,6 +306,25 @@ def test_set_assignee(context_update_example: ActionContext, mocked_jira):
     mocked_jira.set_issue_status.assert_not_called()
 
 
+def test_set_assignee_failing(
+    context_create_example: ActionContext, mocked_jira, caplog
```
grahamalama, line 310: Should we test the `context.operation == Operation.UPDATE` path of `maybe_assign_jira_user` as well to make sure we raise an `IncompleteStepError` in that case?

grahamalama:
By the way

> I'm not entirely sure about the last commit design...

I like the direction you went!

jbi/models.py
leplatrem: Mainly because the API view returns a list of responses. We can change it and make return a map of responses, why not...

tests/unit/actions/test_default.py
leplatrem: I was debugging a broken test and wanted to reduce the scope of executed steps. This is a left over, will revert

jbi/models.py
leplatrem: That was my first attempt, I can't remember why I changed it...

grahamalama (approved):

tests/unit/actions/test_steps.py
grahamalama: Got it -- as long as the path has a test to make that behavior explicit 🙂

PR #579, opened 2023-06-12
Fix #569: documentation whiteboard tags matching

### LINKED ISSUE #569: Document behavior of how JBI parses whiteboard tags
We should explicitly document (in docs files, doctests, etc) how JBI parses whiteboard tags:

- they should be contained in brackets
- JBI will match the whiteboard tag anywhere within a set of brackets as long as it's delimited by dashes
- etc.

leplatrem:
Fix #569

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -226,6 +226,11 @@
   parameters:
     jira_project_key: OMC
     labels_brackets: both
+    issue_type_map:
+      defect: Task
+    status_map:
+      UNCONFIRMED: To Do
+      NEW: To Do
 
 - whiteboard_tag: prodtest
   bugzilla_user_id: 644672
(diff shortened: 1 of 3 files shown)


grahamalama:
I was thinking about this issue as well and thought of a potential bug with the way that we parse whiteboard tags that I opened [here](https://github.com/mozilla/jira-bugzilla-integration/issues/582)

jbi/models.py
```diff
@@ -264,7 +264,12 @@ def extract_from_see_also(self):
         return None
 
     def lookup_action(self, actions: Actions) -> Action:
-        """Find first matching action from bug's whiteboard list"""
+        """
+        Find first matching action from bug's whiteboard list.
```
grahamalama, line 268:
```suggestion
        Find first matching action from bug's whiteboard field.
```

grahamalama (approved):

PR #580, opened 2023-06-12
Fix #578: remove mention of 'allow_private'

### LINKED ISSUE #578: Remove references to private bug syncing in JBI
We previously decided that JBI would _not_ support syncing private bugs. However, there are still aspects of JBI that suggest that private bugs can be synced (the `allow_private` parameter) and we have code paths that account for private bug syncing

```python
if bug.is_private and not action.allow_private:
        f"private bugs are not valid for action {action.whiteboard_tag!r}"
    )
```

We should remove documentation and code paths that relate to private bugs, since JBI does not allow for syncing private bugs.

leplatrem:
Fix #578

### DIFF
--- a/config/config.nonprod.yaml
+++ b/config/config.nonprod.yaml
@@ -8,7 +8,6 @@
     labels_brackets: both
 
 - whiteboard_tag: flowstate
-  allow_private: true
   bugzilla_user_id: 91159
   description: Flowstate whiteboard tag
   parameters:
--- a/jbi/runner.py
+++ b/jbi/runner.py
@@ -44,6 +44,9 @@ def execute_action(
         operation=Operation.HANDLE,
     )
     try:
+        if bug.is_private:
+            raise IgnoreInvalidRequestError("private bugs are not supported")
+
         logger.debug(
             "Handling incoming request",
             extra=runner_context.dict(),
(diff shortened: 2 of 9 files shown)


jbi/runner.py
```diff
@@ -68,7 +68,7 @@ def execute_action(
             ) from err
         runner_context = runner_context.update(action=action)
 
-        if bug.is_private and not action.allow_private:
+        if bug.is_private:
             raise IgnoreInvalidRequestError(
                 f"private bugs are not valid for action {action.whiteboard_tag!r}"
             )
```
grahamalama, line 74: Should we move this guard to the top of this function? We probably don't even want to [try and get the bug](https://github.com/mozilla/jira-bugzilla-integration/blob/dbdbc8bb7bb0b502d761cf4c5d988ffe2fad2bcd/jbi/runner.py#L52) if it's private.

grahamalama: Also, the error message should probably say something about all of JBI not processing private bugs, not just specific actions.

grahamalama (commented):
Nice.

We should probably also remove these references to private bug:
https://github.com/mozilla/jira-bugzilla-integration/blob/dbdbc8bb7bb0b502d761cf4c5d988ffe2fad2bcd/tests/conftest.py#L194 

https://github.com/mozilla/jira-bugzilla-integration/blob/dbdbc8bb7bb0b502d761cf4c5d988ffe2fad2bcd/tests/fixtures/factories.py#L40

https://github.com/mozilla/jira-bugzilla-integration/blob/dbdbc8bb7bb0b502d761cf4c5d988ffe2fad2bcd/jbi/models.py#L177-L180

and maybe I still missed some references.

We should probably also make a note about private comments. Is it possible for a private comment to exist on a public bug? If so, do we want to be syncing that private data?

config/config.nonprod.yaml
```diff
@@ -8,7 +8,6 @@
     labels_brackets: both
 
 - whiteboard_tag: flowstate
```
grahamalama, line 10: We could probably remove this entire config block

grahamalama (approved):

PR #607, opened 2023-06-29
Fix #604: Add product prefix to components for `maybe_update_components`

### LINKED ISSUE #604: Sync Bugzilla components in the `Product::Component` format
From @onemahon in Slack:

> Question about https://github.com/mozilla/jira-bugzilla-integration/issues/518 for [@leplatrem](https://mozilla.slack.com/team/UBJ40EKLL) (I’m sorry, I’m not familiar enough with python to figure this out quickly) - does [maybe_sync_components] support product-scoped components?

> We’re trying to use JBI to sync BZ bugs from multiple products into a single Jira project, but there are some component name collisions (specifically, for example, Fenix::General , Focus::General, and GeckoView::General ).
Does that PR support being able to differentiate based on product? Or would items from all three of those products wind up with the same General component applied in Jira?

leplatrem:
Fix #604

### DIFF
--- a/jbi/services/jira.py
+++ b/jbi/services/jira.py
@@ -197,7 +197,7 @@ def _validate_permissions(all_projects_perms):
 
 def _all_projects_components_exist(actions: Actions):
     components_by_project = {
-        action.parameters.jira_project_key: action.parameters.jira_components
+        action.parameters.jira_project_key: action.parameters.jira_components.set_custom_components
         for action in actions
     }
     success = True
--- a/jbi/steps.py
+++ b/jbi/steps.py
@@ -231,10 +231,15 @@ def maybe_update_components(context: ActionContext, parameters: ActionParams):
     """
     Update the Jira issue components
     """
-    candidate_components = set(parameters.jira_components)
-    if context.bug.component:
+    candidate_components = set(parameters.jira_components.set_custom_components)
+    if context.bug.component and parameters.jira_components.use_bug_component:
         candidate_components.add(context.bug.component)
-    if context.bug.product_component:
+    if context.bug.product and parameters.jira_components.use_bug_product:
+        candidate_components.add(context.bug.product)
+    if (
+        context.bug.product_component
+        and parameters.jira_components.use_bug_component_with_product_prefix
+    ):
         candidate_components.add(context.bug.product_component)
     client = jira.get_client()
 
--- a/jbi/steps.py
+++ b/jbi/steps.py
@@ -234,6 +234,8 @@ def maybe_update_components(context: ActionContext, parameters: ActionParams):
     candidate_components = set(parameters.jira_components)
     if context.bug.component:
         candidate_components.add(context.bug.component)
+    if context.bug.product_component:
+        candidate_components.add(context.bug.product_component)
     client = jira.get_client()
 
     # Fetch all projects components, and match their id by name.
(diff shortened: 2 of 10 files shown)


jbi/steps.py
```diff
@@ -234,6 +234,8 @@ def maybe_update_components(context: ActionContext, parameters: ActionParams):
     candidate_components = set(parameters.jira_components)
     if context.bug.component:
         candidate_components.add(context.bug.component)
+    if context.bug.product_component:
+        candidate_components.add(context.bug.product_component)
```
grahamalama, line 238: I feel like we have a few options when it comes to how we sync a bug's component to an issue. For example take a bug with a product `Product` and a component `General`, as you used in tests below. We could:
- Add the component without the prefix: `General` (what we were doing before this PR)
- Add the component with the product prefix: `Product::General`
- Add both the prefixed and non-prefixed value (what we're doing in this PR)
  - `General`
  - `Product::General`
- Add two components
  - `Product`
  - `General`

Should we give users a choice for how they'd like the component to appear on their issue? If we make our `candidate_components` too "wide" we run the risk of emitting lots of warnings that a component wasn't available in a project. But, an user may only care about the component being added to the issue in a certain way. For example, they may only set up the prefixed notation as valid components on their Jira project, so we'd get an error when trying to set the non-prefixed version.

leplatrem: Very good point. We could have a parameter:

```md
* `use_bugzilla_product_as_jira_component`: `no` / `yes` / `prefix` / `both`
   For a Bugzilla ticket in `Core::General`, use `no` to set component as `General`, use `yes` to set components as `Core` and `General`, use `prefix` to set component as `Core::General`, and `both` to set components as `General` and `Core::General`
```

grahamalama: Yep, some option in the config to control this would be good. I was also thinking something like this?

```yaml
  parameters:
    ...
    jira_components:
      sync_bug_component: <bool>
      sync_product_as_component: <bool>
      sync_bug_component_with_product_prefix: <bool>
      set_custom_components:
      - "my_custom_components"
```

leplatrem: 😎 niice
good idea to use a sub-object

jbi/services/jira.py
```diff
@@ -197,7 +197,7 @@ def _validate_permissions(all_projects_perms):
 
 def _all_projects_components_exist(actions: Actions):
     components_by_project = {
-        action.parameters.jira_project_key: action.parameters.jira_components
+        action.parameters.jira_project_key: action.parameters.jira_components.set_custom_components
```
grahamalama, line 200: Not something to address in this PR, but:

Looking at this check again, it seems a little less valuable considering we can only check that the _custom_ components are set in the heartbeat. We don't know if all of the possible components that come via a bug will be set in the project.

I wonder if it's possible and if it would be beneficial to give the Jira automation bot the permissions to add components to projects when they don't exist. That way, we could add the component to the project before setting it on the issue if a certain component didn't exist.

grahamalama (approved):

jbi/services/jira.py
leplatrem: > I wonder if it's possible and if it would be beneficial to give the Jira automation bot the permissions to add components to projects when they don't exist. That way, we could add the component to the project before setting it on the issue if a certain component didn't exist.

Yes, I don't think this permission could be harmful anyway...

PR #612, opened 2023-06-30
Make all factories fixtures

grahamalama:
This PR:

- sets up all factories as pytest fixtures
- replaces the direct use of factories in tests with their fixture counterparts
- where possible, uses factories instead of patching built fixtures
For example instead of doing 

```python
foo = foo_example.bar.baz = "override"
```
do

```python
foo = foo_factory(bar=bar_factory(baz="override"))
```

While this is a little more verbose, it makes the data that's being generated more obvious and consistent between different tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -108,46 +108,6 @@ def context_create_example(action_context_factory) -> ActionContext:
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


leplatrem (approved):
LGTM

PR #615, opened 2023-07-04
Prefix component with product on FXP and SP3 projects

davehunt:
Both projects have the components field on the create/update screens. This sync is already working, but I'd like to include the product as we have some generic/shared components.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -130,6 +130,9 @@
   description: Performance Team
   parameters:
     jira_project_key: FXP
+    jira_components:
+      use_bug_component: false
+      use_bug_component_with_product_prefix: true
     steps:
       new:
         - create_issue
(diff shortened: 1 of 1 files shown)


leplatrem (approved):
LGTM

With this config, if both `Core::General` and `General` components exist, they will be both assigned (`use_bug_component` is true by default)

davehunt:
> With this config, if both `Core::General` and `General` components exist, they will be both assigned (`use_bug_component` is true by default)

Ah, I was a little confused by the ability to set multiple components. So how do I make it so only `Core::General` would be set? Would it be as follows?

```
jira_components:
  use_bug_component: false
  use_bug_component_with_product_prefix: true
```

davehunt:
@leplatrem please could you take another look?

leplatrem (approved):
Great 💯 

(It may be a sign that this setting does not have the right default value)

Micha3L2020 (approved):

PR #619, opened 2023-07-07
Validate that step functions exist when parsing actions

### LINKED ISSUE #587: Validate that all mentioned steps in config are known Python functions
See #585

grahamalama:
Closes #587


jbi/models.py
```diff
@@ -22,6 +22,13 @@
 JIRA_HOSTNAMES = ("jira", "atlassian")
 
 
+@functools.lru_cache
+def steps_module_has_function(function_name):
+    from jbi import steps
```
grahamalama, line 27: I don't love that I had to do this late import (it's usually a code smell to me), but without this, I was running into circular imports.

jbi/models.py
```diff
@@ -41,6 +48,13 @@ class ActionSteps(BaseModel):
         "create_comment",
     ]
 
+    @validator("*")
+    def validate_steps(cls, function_names):
+        for function_name in function_names:
+            if not steps_module_has_function(function_name):
+                raise ValueError(f"{function_name} is not a valid step function")
```
leplatrem, line 55: nitpick: raise a valuerror with all invalid step names (in order to avoid several loops of config change / fix / redeploy)

jbi/models.py
leplatrem: I think this could be acceptable in this particular case since we are introspecting modules, but yeah I see what you mean

leplatrem (approved):

PR #621, opened 2023-07-10
Reduce coupling between services and steps modules

grahamalama:
This PR was partially motivated by #619. In it, importing the `steps` module into the `models` module to introspect it caused a circular import error.

The simple way to fix that would have been to do a late import. But, that being a code smell, I wondered if there were other ways to solve the problem and make other improvements in the process. This PR represents one attempt. It:

- Slightly refactors and clarifies the purposes of the classes found in the `services` modules. 

  Now, each module has two classes, a `Client` and a `Service`. The client class's only responsibility should be to make, receive, and parse HTTP requests. It then is used by a `Service` to complete "high-level" operations.

  This change helps clarify the responsibilities of each class. It also makes things slightly easier to test, in my opinion. Now, we can rely less on patching and instead pass fake or mock clients to services to make assertions.

- Reduces coupling between the `services` and `steps` modules

  Instead of implicitly depending on the services modules, step functions now take services as arguments. This allows us to not have to import the services directly into the steps module (only for type checking) and again makes testing easier by allowing us to make mocks or fakes as arguments rather than by patching.

- Makes the services FastAPI dependencies for router functions

After this change (in a follow up PR), importing the `steps` module into the `models` module to introspect it causes no errors.

This is a large change for a relatively small goal. If we decide we're okay with one late import, I can close this PR in favor of the "simple" approach.

Also, one problem lurking in the shadows is that as soon as we have to import a model into the `steps` module, we'll run into this problem again. Hopefully it doesn't come to that as long as we can use models only for typing purposes.

### DIFF
--- a/jbi/router.py
+++ b/jbi/router.py
@@ -18,6 +18,7 @@ from jbi.services import bugzilla, jira
 SettingsDep = Annotated[Settings, Depends(get_settings)]
 ActionsDep = Annotated[Actions, Depends(get_actions)]
 VersionDep = Annotated[dict, Depends(get_version)]
+BugzillaServiceDep = Annotated[bugzilla.BugzillaService, Depends(bugzilla.get_service)]
 router = APIRouter()
 
 
@@ -38,10 +39,12 @@ def root(request: Request, settings: SettingsDep):
 
 @router.get("/__heartbeat__")
 @router.head("/__heartbeat__")
-def heartbeat(response: Response, actions: ActionsDep):
+def heartbeat(
+    response: Response, actions: ActionsDep, bugzilla_service: BugzillaServiceDep
+):
     """Return status of backing services, as required by Dockerflow."""
     health_map = {
-        "bugzilla": bugzilla.check_health(),
+        "bugzilla": bugzilla_service.check_health(),
         "jira": jira.check_health(actions),
     }
     health_checks: list = []
@@ -92,9 +95,9 @@ def get_whiteboard_tags(
 
 
 @router.get("/bugzilla_webhooks/")
-def get_bugzilla_webhooks():
+def get_bugzilla_webhooks(bugzilla_service: BugzillaServiceDep):
     """API for viewing webhooks details"""
-    return bugzilla.get_client().list_webhooks()
+    return bugzilla_service.client.list_webhooks()
 
 
 @router.get("/jira_projects/")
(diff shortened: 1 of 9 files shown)


jbi/runner.py
```diff
@@ -69,14 +81,31 @@ def _initialize_steps(self, steps: ActionSteps):
...
         return steps_callables
 
+    def build_step_kwargs(self, func) -> dict:
+        """Builds a dictionary of keyword arguments (kwargs) to be passed to the given `step` function.
+
+        Args:
+            func: The step function for which the kwargs are being built.
+
+        Returns:
+            A dictionary containing the kwargs that match the parameters of the function.
```
leplatrem, line 91: Do we have a convention already to use Google's docstring format?

jbi/steps.py
```diff
@@ -28,22 +32,30 @@ def create_comment(context: ActionContext, parameters: ActionParams):
...
+    bugzilla_service: BugzillaService,
+):
     """Create the Jira issue with the first comment as the description."""
     bug = context.bug
     issue_type = parameters.issue_type_map.get(bug.type or "", "Task")
 
     # In the payload of a bug creation, the `comment` field is `null`.
     # We fetch the list of comments to use the first one as the Jira issue description.
-    comment_list = bugzilla.get_client().get_comments(bug.id)
+    comment_list = bugzilla_service.client.get_comments(bug.id)
```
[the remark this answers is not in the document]
grahamalama: Yeah, this is a good idea. I made some changes to avoid direct client access here [653b4a4](https://github.com/mozilla/jira-bugzilla-integration/pull/621/commits/653b4a4c6afc5b7262097cb72b7ed35be7da158e)

Probably worth another review.

jbi/router.py
```diff
@@ -92,15 +99,15 @@ def get_whiteboard_tags(
 
 
 @router.get("/bugzilla_webhooks/")
-def get_bugzilla_webhooks():
+def get_bugzilla_webhooks(bugzilla_service: BugzillaServiceDep):
     """API for viewing webhooks details"""
-    return bugzilla.get_client().list_webhooks()
+    return bugzilla_service.client.list_webhooks()
```
leplatrem, line 104: Leftover?

jbi/steps.py
```diff
@@ -3,21 +3,25 @@
 
 Each step takes an `ActionContext` and a list of arbitrary parameters.
 """
+from __future__ import annotations
```
leplatrem, line 6: Can you explain why do we need this here? I thought importing everything necessary on top with `if TYPE_CHECKING` was enough

Maybe add a link to https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations

leplatrem (approved):

PR #636, opened 2023-07-26
Fix #635: ignore update workflow when project does not match config

### LINKED ISSUE #635: Errors occur when linked Jira issue does not match project in config
https://mozilla.sentry.io/issues/4320816876/?referrer=slack&alert_rule_id=11827252&alert_type=issue

**Steps to reproduce**

1. configure tag `[foo]` to be associated with project `FOO`
2. `bug 9999` receives tag `[foo]`
3. Jira issue `FOO-123` gets created
4. `bug 9999` receives link to `FOO-123` 
5. configuration is changed to associate `[foo]` with project `BAR`
6. `bug 9999` receives a change (eg. status)
7. actions associated to `[foo]` are executed with project `BAR` in context
8.  actions like `maybe_update_components` will inspect the `BAR` project, and will try to apply the changes on the `FOO-123` issue which is in the `FOO` project.

**Solution Ideas**

* Abort workflow when linked issue does not match project in config
* Put the project from the linked issue as context instead of the one in config

The former makes more sense, because for other actions like `maybe_update_status` where a mapping is defined in configuration, using the linked issue's project can be confusing since it may not have the same mappings...

leplatrem:


grahamalama (approved):

leplatrem:
Could verify that field is provided in GET response:

```
$ curl --request GET \
  --url 'https://mozilla-hub.atlassian.net/rest/api/2/issue/FNXV2-263' \
  --user 'mleplatre@mozilla.com:$JIRA_KEY_PASS' \
  --header 'Accept: application/json' | jq .fields.project.key

"FNXV2"
```

PR #642, opened 2023-07-27
Add PCF config for PerfCompare project

davehunt:


leplatrem:
Could you please double check that:
- the Jira automation user has the `DELETE_ISSUES,ADD_COMMENTS,EDIT_ISSUES,CREATE_ISSUES` permissions on the PCF project
- status and resolution fields are on the create and update screens
- you don't want to synchronize components

davehunt:
> Could you please double check that:
> 
>     * the Jira automation user has the `DELETE_ISSUES,ADD_COMMENTS,EDIT_ISSUES,CREATE_ISSUES` permissions on the PCF project

I have granted jira-automation@mozilla.com the Bots role.

>     * status and resolution fields are on the create and update screens

I can confirm that the status field is present. The resolution field is not, but it's also not present on the SP3 project, where JBI is working.

>     * you don't want to synchronize components

That's correct. All of these bugs will be in the same bugzilla component.

leplatrem (approved):

PR #701, opened 2023-10-10
Use tag and abbrev as version in container and Sentry

leplatrem:
This aligns JBI with Remote Settings and CTMS


.github/workflows/build-publish.yaml
```diff
@@ -22,9 +22,10 @@ jobs:
 
       - name: Build `version.json` file
         run: |
```
bsieber-mozilla, line 24: I believe this needs to be move up/under the "env" section

leplatrem:
Screenshot 2023-10-10 at 19 02 09

Issue #724, opened 2023-10-26
Application fails to parse `SENTRY_DSN` environment variable


grahamalama:
https://mozilla.slack.com/archives/C019WG3TTM2/p1698344672467099

<img width="1212" alt="image" src="https://github.com/mozilla/jira-bugzilla-integration/assets/20747904/71d2574c-5621-4bc3-a8b3-f571e93adf9c">

[Logs Explorer link](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22moz-fx-jbi-nonprod-7e35%22%0Aresource.labels.location%3D%22us-west1%22%0Aresource.labels.cluster_name%3D%22jbi-nonprod-v1%22%0Aresource.labels.namespace_name%3D%22stage-jbi%22%0Aresource.labels.pod_name%3D%22jbi-stage-jbi-app-1-canary-7f96c6b758-2dh7d%22%20severity%3E%3DDEFAULT%0Aresource.labels.container_name%3D%22jbi%22;cursorTimestamp=2023-10-26T17:07:48.488013054Z;startTime=2023-10-25T17:12:51.400Z;endTime=2023-10-26T17:12:51.400Z?project=moz-fx-jbi-nonprod-7e35)

PR #725, opened 2023-10-26
Fix Sentry url parsing

grahamalama:
Fixes: #724

### DIFF
--- a/jbi/environment.py
+++ b/jbi/environment.py
@@ -19,10 +19,6 @@ class Environment(StrEnum):
     PROD = "prod"
 
 
-class SentryDsn(AnyUrl):
-    """Url type to validate Sentry DSN"""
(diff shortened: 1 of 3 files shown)


jbi/environment.py
```diff
@@ -51,7 +43,7 @@ class Settings(BaseSettings):
     log_format: str = "json"  # set to "text" for human-readable logs
 
     # Sentry
-    sentry_dsn: Optional[SentryDsn] = None
+    sentry_dsn: Optional[AnyUrl] = None
```
grahamalama, line 46: I don't know _why_ this was the fix. I'm guessing this was fallout of upgrading to Pydantic v2.

leplatrem (approved):
:100: :shipit:

Issue #334, opened 2022-11-28
Issue with Resolution Maps


bsieber-mozilla:
@bsieber-mozilla Do you want to leave this issue open to fix the resolution_map problem? Or close this issue and file a new issue for resolution_map?

_Originally posted by @cpeterso in https://github.com/mozilla/jira-bugzilla-integration/issues/301#issuecomment-1325703755_

bsieber-mozilla:
Closed in favor of-- https://github.com/mozilla/jira-bugzilla-integration/issues/294

PR #735, opened 2023-11-01
Add resolution map for FXP, PCF, and SP3

davehunt:
When I originally tried to include a resolution map there seemed to be some issues, which led to it being removed. I think it was related to #334 but it's not clear to me now so I thought I'd try again. If it's still not working as documented, maybe we can try to work through the issues? In preparation for this change I have made sure that the resolution field is present for all of the projects, and that all resolution values mentioned in this patch are available in Jira.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -390,6 +390,15 @@
       WORKSFORME: Done
       INCOMPLETE: Done
       MOVED: Done
+    resolution_map:
+      FIXED: Done
+      INVALID: Invalid
+      WONTFIX: "Won't Do"
+      INACTIVE: Inactive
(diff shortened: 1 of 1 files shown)


leplatrem:
If you made sure the resolution values are present, we should be good to go!

leplatrem (approved):

leplatrem:
@davehunt We had to revert this PR because of this weird error:
```
data='{"fields": {"resolution": "Duplicate"}}' 

--> error={"resolution": "Specify the Resolution (id or name) in the string format"}
```

We opened #750 but maybe you can re-open a PR so that we don't loose track of your desired changes?

davehunt:
I opened #751

PR #736, opened 2023-11-01
Add labeler action

grahamalama:
When we closed https://github.com/mozilla/jira-bugzilla-integration/pull/624, we didn't actually add a workflow to call the [actions/labeler](https://github.com/actions/labeler) action 🤦🏻 

https://github.com/actions/labeler#create-workflow


leplatrem (approved):
Should work better now ;) Thank you!

Issue #645, opened 2023-07-28
Assess the state of our logging


grahamalama:
We realized in #644 that we weren't logging some valuable information at the correct log level. #644 was a "quick fix", but we should go through our app and examine all the logging we're doing to answer these questions:

- are we logging messages at the correct level?
- are we logging the right amount information?

leplatrem:
Something I noticed while investigating #747 is that our `logger.debug()` messages don't show up in Log Explorer. Maybe we should log most of them at INFO level 🤷

PR #767, opened 2023-11-17
Start process to improve Jira logging

grahamalama:
Related #645 

Adds / modifies logging for a few Jira service methods

- get_issue
- create_issue
- update_issue_resolution

The important things we want to check here:
- things are logged at the appropriate level (Debug)
- we include the `rid` everywhere, as this is what will help us correlate logs for an action workflow
- We test this in a sane way

This PR is getting a temperature-check for the logging / testing patterns before I continue further


jbi/services/jira.py
```diff
@@ -253,33 +261,49 @@ def create_jira_issue(
...
+                message = ", ".join(str(exc).split("\n"))
+                raise JiraCreateError(message) from exc
+            except json.JSONDecodeError:
+                # If the response is not JSON, use the response text
+                logger.exception(
+                    "Failed to create issue for Bug %s",
+                    bug.id,
+                    extra={"response": exc.response.text, **context.model_dump()},
+                )
+                raise JiraCreateError(str(exc)) from exc
```
leplatrem, line 295: Wouldn't this be more concise? Is the `join`/`split` specific to the http exception?

```suggestion
            try:
                response = exc.response.json()
            except json.JSONDecodeError:
                # If the response is not JSON, use the response text
                response = exc.response.text
            
            logger.exception(
                "Failed to create issue for Bug %s",
                bug.id,
                extra={"response": response, **context.model_dump()},
            )
            message = ", ".join(str(exc).split("\n"))
            raise JiraCreateError(message) from exc
```

tests/unit/services/test_jira.py
```diff
@@ -155,3 +154,231 @@ def test_all_projects_components_exist_no_components_param(
...
+
+    before, after = capturelogs.records
+    assert before.levelno == logging.DEBUG
+    assert before.message == "Getting issue JBI-234"
+
+    assert after.levelno == logging.ERROR
+    assert after.message.startswith("Could not read issue JBI-234")
+
+
+def test_get_issue_raises_other_error(
```
leplatrem, line 199: nit: other error*s* (?)


Below you have just `test_update_..._raises(` which is fine too :)

tests/unit/services/test_jira.py
```diff
@@ -155,3 +154,231 @@ def test_all_projects_components_exist_no_components_param(
...
+
+    before, after = capturelogs.records
+    assert before.message == f"Creating new Jira issue for Bug {context.bug.id}"
+    assert before.fields == issue_fields
+
+    assert after.message == f"Jira issue JBI-234 created for Bug {context.bug.id}"
+    assert after.response == mocked_response_data
+
+
+def test_create_jira_issue_returns_list(
```
leplatrem, line 308:
```suggestion
def test_create_jira_issue_when_list_is_returned(
```
?

leplatrem (approved):
👌 
(you can ignore any of my comments if they don't make sense to you)

jbi/services/jira.py
grahamalama: Good call. I changed things up a big more in [e7aa369](https://github.com/mozilla/jira-bugzilla-integration/pull/767/commits/e7aa36959f50b8d17602e9fa42f0c1e3759bee72) -- lmk what you think

jbi/services/jira.py
```diff
@@ -253,33 +261,49 @@ def create_jira_issue(
...
-        jira_response_create = self.client.create_issue(fields=fields)
+        logger.debug(
+            "Creating new Jira issue for Bug %s",
+            bug.id,
+            extra={"fields": fields, **context.model_dump()},
+        )
+        try:
+            response = self.client.create_issue(fields=fields)
+        except requests.HTTPError as exc:
+            assert exc.response is not None
```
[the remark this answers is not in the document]
grahamalama: True. This was just to make mypy happy -- we can be reasonably sure we'll have a response object here, no?

If we continue to fight with `mypy` about safely accessing `exc.response`, maybe it deserves a helper 🤔

tests/unit/services/test_jira.py
grahamalama: `nit: other errors (?)`

Maybe "raises" is the wrong word here -- my goal was to illustrate that we the Jira Client calls raise_for_status, we swallow `404`s, but reraise everything else. I'll change the name of this test to "test_get_issue_reraises_other_erroring_status_codes"

Issue #743, opened 2023-11-03
Create alerts for when bugs fail to sync


grahamalama:
Configure Grafana or Sentry alerts when:
- [ ] We fail to sync a bug at all
- [ ] We only partially sync a bug (e.g. we sync whiteboard labels, but can't sync status)

leplatrem:
When the payload has failed to be processed several time. Before it expires, we could also consider to alert by trying (in order):
- to send a comment on bugzilla
- to send a comment on Jira
- fetch the email of the bugzilla user id associated to the project, and send them an email

alexcottner:
@grahamalama - would you consider this resolved with our IM alerts? Or are you looking to have historical data we can report on?

Issue #730, opened 2023-10-31
Establish convention for capturing system incidents


grahamalama:
We are trying to reduce the frequency of incidents which cause customers to ask, "why isn't my bug syncing?" In order answer the question "are we reducing the frequency?", we need to capture this data consistently.

In an ADR, we should establish a workflow for tracking incidents, including the start time and resolution time. This way, we can capture both the number of incidents and the average time to resolve incidents.

Issue #710, opened 2023-10-15
System is able to "drop" erring messages without user interference


bsieber-mozilla:
```Yeah, I wonder if part of the answer here is to remove some of the burden from the heartbeat / SRE by making the system either more self-healing or self-repairable by customers.``` -- https://github.com/mozilla/jira-bugzilla-integration/issues/343#issuecomment-1355150969

Along with additional validation features, like config-validation prior to merge, the goal would be for errors to not "halt" processing for any users. 

When an error causes a WH queue to no longer send messages, the system becomes dependent on an admin to drop the erring message. 

This act of dropping the message could be automated when a bugzilla request is received--if the processing results in a failure or error, that should be caught--once caught if the amount of webhook errors (from the webhook api) is greater than X (0?) than the message could be "dropped". This would align with current human-workflows.

_We could possibly further this process by making this action transparent; via a slack message to the channel with the associated bug id._

- [ ] System is able to "drop" erring messages without user interference

grahamalama:
As we discussed on Zoom:

In addition to dropping messages, we could (should?) also try to find a way to drop issues that are blocking the queue through the Bugzilla API. This endpoint doesn't currently exit -- we'd have to make this request to the Bugzilla devs to help us out.

bsieber-mozilla:
> As we discussed on Zoom:
> 
> In addition to dropping messages, we could (should?) also try to find a way to drop issues that are blocking the queue through the Bugzilla API. This endpoint doesn't currently exit -- we'd have to make this request to the Bugzilla devs to help us out.

To "drop" a message could be to send a 200 back to bugzilla when the errors in the webhook queue api is greater than 0 and at the end of processing a request it would've resulted in an error. 

Perhaps its a bit odd to say 200/Success for an erring message--but I see this 200 as more of a, we've processed this message in a way that's aligned with how our system should process it. Any side-effects from the request being not sync'd (error messages on bugzilla bug) could be done as a last step in the processing?

In terms of using the Bugzilla API; it would be nice to have the feature mentioned:
- drop last/latest request
- read error for last/latest request
- enable/disable webhook
- ...

grahamalama:
> To "drop" a message could be to send a 200 back to bugzilla when the errors in the webhook queue api is greater than 0 and at the end of processing a request it would've resulted in an error.

That's true and is a good defensive option. It doesn't _completely_ save us, but it would be a good addition to also having the ability to clear the queue the queue through the API.

[...]

PR #775, opened 2023-11-27
ADR - Options to prevent bugzilla from pausing webhook calls

alexcottner:
Added ADR with proposed options to prevent bugzilla from pausing webhook calls due to configuration errors.

[Link to ADR for easy reference](https://github.com/alexcottner/jira-bugzilla-integration/blob/adr-003/docs/adrs/003-bugzilla-response-codes.md).

Issues this could solve (or affect) depending on solution we implement:
- https://github.com/mozilla/jira-bugzilla-integration/issues/710
- https://github.com/mozilla/jira-bugzilla-integration/issues/743
- https://github.com/mozilla/jira-bugzilla-integration/issues/730


bsieber-mozilla:
Thanks for this ADR @alexcottner ! This is certainly an issue worthy of an ADR. In terms of options--I think option 1, possibly with some slight addition would be a good approach.

In terms of options 3, 4, 5: I'm hesitant for adding additional infra, like a redis-cache/queue, when there's already a "queue" in the system (the bugzilla webhook queue). I understand it would give JBI more control over the messages, but also is data redundant.

In terms of option 1, depending on how we log, we could use those logs to automate alerts (https://cloud.google.com/logging/docs/alerting/log-based-alerts) which could then be sent to slack/email/other channels. 

That way we "log and move-on" but also have a breadcrumb trail in public space so our users could be aware that they might be affected by data-loss. The alert could reveal the particular configuration, and bug_id so that interested parties could jump if they were expecting something that did not appear. 

Thanks again for spinning up this ADR! 🙇

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Decision Drivers
+
+- Amount of initial engineering effort
+- Amount of maintenance effort
+- Overall performance of JBI (how quickly is data able to move)
+- How intuitive the solution is to the users that depend on the data (will picking the easiest option solve their needs?)
+
+## Considered Options
+For all of these options, we will be returning a succesful response to bugzilla's webhook calls. I propose a 202 status code to identify that we're accepting the request immediately and are processing it asynchronously.
```
leplatrem, line 21:
```suggestion
For all of these options, we will be returning a successful response to bugzilla's webhook calls. I propose a 202 status code to identify that we're accepting the request immediately and are processing it asynchronously.
```

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Option 1: Log the failure and move on
+JBI will log that we couldn't process a specific payload, along with relevant ID's (bug id, jira ticket id, comment id, etc) so further investigation can be done if needed.
+
+Pros:
+- The simplest solution
+- Very low effort
+
+Cons: 
+- Will not alert people to data loss (without additional alerting functionality)
```
leplatrem, line 31: It could try to add a comment about it on the bugzilla ticket and/or jira issue

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+- Still requires engineers to investigate further
+
+### Option 2: Ask a human to do something
+JBI will alert users that data could not be sync'd. This could happen through an IM alert or an email immediately, or a scheduled (daily?) report.
+
+### Option 3: Queue retries internally
+Create a persistence layer within the JBI containers that will queue and retry jobs for a specific length (2 hours?) of time. This could be done with an internal cache (redis) or database (postgres) within the container. After retries exceed the max time length, an error would be logged and the data would be dropped.
```
leplatrem, line 38: what do you mean _within the container_? Like in memory?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Option 4: Queue data externally
+Similar to option 3, but instead of JBI queing the events we would have a dedicated service that accepts all API calls from bugzilla and puts them into a queue. JBI would shift to being a downstream service and process these events asynchronously. If events fail to process to many times, they would get sent to a DLQ (dead letter queue) that could be replayed later if needed.
+
+There are plenty of existing solutions we could use to solve this problem from a technical perspective. A seperate ADR would be done to identify the best answer if we choose to go this route.
+
+## Option 5: A combination of the above
+Example: We could create an external queue for processing and then alert users if the DLQ grows to quickly..
```
leplatrem, line 46:
```suggestion
Example: We could create an external queue for processing and then alert users if the DLQ grows too quickly..
```

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+When bugzilla receives too many error responses from JBI, it stops triggering webhook calls for the entire project causing data to stop syncing. Frequently these errors are due to JBI being unable to process a payload due to errors in configuration (or incomplete configuration) in Jira or a mismatch of data for a single bug. 
+
+We don't want the entire sync process to stop because of this. We have identified four options to solve this problem.
+
+## Decision Drivers
+
+- Amount of initial engineering effort
+- Amount of maintenance effort
+- Overall performance of JBI (how quickly is data able to move)
+- How intuitive the solution is to the users that depend on the data (will picking the easiest option solve their needs?)
```
leplatrem, line 18: I don't find mention of these decision drivers in the pros and cons paragraphs below.  A suggestion would be to give them a title: (eg. `Engineering Efforts`) and then below list them again: `Engineering Efforts: Good because X`

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+
+Pending discussion
+
+### Negative Consequences <!-- optional -->
+
+Pending discussion
+
+## Pros and Cons of the Options <!-- optional -->
+
+### Option 1: Log the failure and move on
```
leplatrem, line 63: Why don't you merge the above paragraphs that describe the option with these ones that list pros and cons?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+- The simplest solution
+- Very low effort
+
+Cons: 
+- Will not alert people to data loss (without additional alerting functionality)
+- Still requires engineers to investigate further
+
+### Option 2: Ask a human to do something
+Pros:
+- Removes need for engineering
```
leplatrem, line 74: Depending on who is going to receive the alert. 
Will it be a Jira admin? A project owner? 
Will they be able to resolve the issue themselves or will they open a issue on JBI's Github?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
...
+- Alerts users directly that there is a problem
+- Moderate effort
+
+Cons:
+- Alerts can be noisy and cause notification fatigue
+
+### Option 3: Queue retries internally
+Pros:
+- Allows for retries up to a designated amount of time
+- Keeping all services within the container make dev work easier compared to an external queue
```
leplatrem, line 84: How does it work when using multiple web workers?

Dev work easier, like in not having to use docker-compose? I don't understand

docs/adrs/003-bugzilla-response-codes.md
leplatrem, line 38: Also, usually the issues are due to a Jira/JBI misconfiguration, and it is very unlikely that it will be resolved within a few hours. Sometimes it even takes days because we depend on Jira admins perms that we don't have

leplatrem (commented):
See https://github.com/mozilla/jira-bugzilla-integration/issues/82

One main comment: the format would require a bit more structure to follow the arguments better (criteria/decision drivers in front of each proposed solution). See other ADRs in docs/adrs/

Also, there is another solution that Graham suggested a while ago: when onboarding a project on JBI,  we assign a stakeholder that is in charge of setting up the webhook for their product in Bugzilla. When there is an error, this stakeholder is alerted, and is charge of fixing it to unblock their queue.

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
+# Bugzilla disables webhooks after too many errors
+
+- Status: proposed
+- Date: 2023-27-11
+
+Tracking issue:
```
leplatrem, line 6: https://github.com/mozilla/jira-bugzilla-integration/issues/82
https://github.com/mozilla/jira-bugzilla-integration/issues/181
https://github.com/mozilla/jira-bugzilla-integration/issues/21

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,100 @@
+# Bugzilla disables webhooks after too many errors
+
+- Status: proposed
+- Date: 2023-27-11
+
+Tracking issue: 
+
+## Context and Problem Statement
+When bugzilla receives too many error responses from JBI, it stops triggering webhook calls for the entire project causing data to stop syncing. Frequently these errors are due to JBI being unable to process a payload due to errors in configuration (or incomplete configuration) in Jira or a mismatch of data for a single bug.
```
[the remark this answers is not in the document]
grahamalama: > If we configure a webhook per product, it will only stop sending events for this product and won't affect the rest (See https://github.com/mozilla/jira-bugzilla-integration/issues/21).

This is the way it is right now in production -- we have many webhooks set up for each Bugzilla product (scoped to `ANY` component).

image

grahamalama, line 9: The most recent incidents have been because the webhook payload doesn't match our schemas in some way. FastAPI will return `422` errors for Pydantic validation errors.

docs/adrs/003-bugzilla-response-codes.md
grahamalama, line 21: We might be limited by Bugzilla here. ([docs](https://bugzilla.mozilla.org/page.cgi?id=webhooks.html))

> If the response is not 200, the Bugzilla system will retry 4 attempts, waiting 5s, 10s, 15s and 20s between each attempt. If none of those attempts is successful, the system will continue trying every 15 minutes.

docs/adrs/003-bugzilla-response-codes.md
grahamalama: @leplatrem I think your suggestion might be covered in Option 2. This seems most like a "do nothing" option?

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Will add clarifying language.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Dev work becomes much easier for end-to-end testing and local debugging. Will expand on that more. When you're working off of external queues you can't to e2e tests until you deploy somewhere. Unless you want to run a bunch of containers locally, but then you also run into weird compatibility issues between operating systems and all sorts of fun stuff.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: I was really hoping they just checked if the response status is OK, but you're right they're intentionally checking for 200. Will update the ADR.
https://github.com/bugzilla/harmony/blob/e72338d1a8c374e0f1ee88770a05701067fa6b64/extensions/Push/lib/Connector/Webhook.pm#L113

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Added Decision Drivers section under each option.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: As in: running a container that has JBI and a cache or DB within it for retries. We could also run another separate container to host this layer, but at that point why not just run a dedicated queue service (which would be option 4)?

leplatrem:
About option 1: log and move on. How long do we retain logs for? 

What about a solution where we log the error (+notification to devs on Slack) and have a super simple DLQ? 
For example, we would append failing webhook payloads to a file, and have a command to replay them and purge the file? Payloads failing again during retry would be added to DLQ of the current day/month.
If file-based sounds too prehistoric and not flexible enough for statistics and stuff, we could have a simple bigquery table, that unlike logs could be kept forever.

alexcottner:
> About option 1: log and move on. How long do we retain logs for?

Currently this is 30 days in log-explorer and 90 days in bigquery. But we could do this for as long or as little as we wanted, or even stream these specific logs to a bucket as a DLQ.

> What about a solution where we log the error (+notification to devs on Slack) and have a super simple DLQ? For example, we would append failing webhook payloads to a file, and have a command to replay them and purge the file? Payloads failing again during retry would be added to DLQ of the current day/month. If file-based sounds too prehistoric and not flexible enough for statistics and stuff, we could have a simple bigquery table, that unlike logs could be kept forever.

Appending to _single_ a file is annoying when running parallel containers. But if the spirit of this is "send it to a DLQ that we could re-process later", then it's easy enough to send them to a bucket or a volume on K8s. It would probably be something like `/{year}/{month}/{date}/{pod-id}-{message-id}.json`.

alexcottner:
We had a discussion today where I think we're leaning towards one or both of these:

- Option 2, ask a human to do something
- Option 4, use a simple DLQ

Both of these are relatively small efforts to implement with good outcomes. If we can easily generate alerts based on our simple DLQ (ex: an emailed report from a postgres table or messaging an IM channel based on a file being dropped into a bucket) then this could be a simple solution with high resiliency.

leplatrem:
> Both of these are relatively small efforts to implement with good outcomes. If we can easily generate alerts based on our simple DLQ (ex: an emailed report from a postgres table or messaging an IM channel based on a file being dropped into a bucket) then this could be a simple solution with high resiliency.

💯 

I vote to start with _Option 4_ with files on disk (mounted volume), and iterate if needed

And with regards to alerting, we could have another dedicated ADR (bugzilla comment vs Slack vs email to workflow owner, etc.)

grahamalama:
I think we're underestimating the engineering effort of Option 4. 

> We would always return 200, but any events that fail to process internally would get sent to a DLQ and be replayed later if needed. ... A scheduled kubernetes job would then run to try and pick these up and reprocess them later (every 4 hours, for exmaple).

There's a lot packed in here. How will we decide when it's needed (or not needed) to replay a message from the DLQ? How will we avoid situations where, for example:

- user creates a bug, it syncs
- user adds `EDIT 1: ...` to the summary
- some network error happens, the bug fails to sync, and it's placed in the DLQ
- the network error solves itself
- user adds `EDIT 2: ...` to the summary. The summary is synced.
- 4 hours later, we replay the message in the DLQ. The summary is synced, but it erases `EDIT 2` since it wasn't on the message in the DLQ


Also, something I saw on [AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html#sqs-dead-letter-queues-when-to-use) on DLQs:

> Don't use a dead-letter queue with a FIFO queue if you don't want to break the exact order of messages or operations.

I think for now, that's us (illustrated by the example above). I think we _could_ code our way around it in certain cases, but that adds to the complexity of this option

grahamalama:
Also, something I noticed in the ADR:

```
Option 2:
- Amount of initial engineering effort: low
- Amount of maintenance effort: low
```

```
Option 4:
- Amount of initial engineering effort: mid
- Amount of maintenance effort: mid-low
```

Based on our own criteria, Option 2 requires lower effort and maintenance that Option 4. Performance and intuitiveness are the same. What then, is compelling us to choose Option 4? Are there some other considerations that we haven't documented in the ADR that make Option 4 more attractive?

alexcottner:
Good points, processing data out of order can be just as dropping it. Gonna update the ADR with that as a risk for now.

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+  <summary>Breakdown of flowchart</summary>
+
+  1. JBI receives a payload from Bugzilla or the Retry Scheduler.
+  1. JBI will always return 200/OK for a response.
+  1. If the bug is private, discard the event and log why.
+  1. If the bug cannot be found in the Bugzilla API, discard the event and log why.
+  1. If an associated action cannot be found for the event, discard the event and log why.
+  1. If a matching Jira issue cannot be found, and the event is not creating one, discard the event and log why.
+  1. If there is a mismatch between project keys in the event and Jira, discard the event and log why.
+  1. If there is a blocking event in the DLQ, skip to the Error Event Handler.
```
leplatrem, line 43: How do you define a *blocking* event?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+  1. If there is a merge-able event in the DLQ, merge with the current event (current event wins conflicts).
```
leplatrem, line 44: What makes an event *merge-able*? Same type?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+  1. Write updated data to Jira's API.
+      1. If successful, delete any associated items in DLQ.
```
leplatrem, line 46: Don't we always want to delete the items from DLQ? I mean, if it fails again, it will be added again no?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+  1. Handle errors in Error Event Handler
+      1. Write an error to the logs, which may be forwarded to an alerting mechanism.
+      1. Write an updated event file to the DLQ if the original event is less than 7 days old.
+      1. If we have exceeded 7 days from the original event, delete associated DLQ items.
+  1. The retry scheduler runs every 12 hours and will re-send events to JBI. Starting with potentially blocking events first.
+</details>
+
+### Pros:
+ - Avoids the problem of accidentally overwriting newer data with older data
+ - Avoids making users correct data manually if something is misconfigured
```
leplatrem, line 57: I didn't get that one 🤔 
How does it prevent users from correcting? Correcting what? Fields on Jira?

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+ - Gives users a whole work week to update potentially misconfigured settings
+ - Low maintenance effort
+ - Mid-low engineering effort
+ - High performance of JBI
+ - Intuitive solution with alerting via error logs
+
+### Cons:
+ - Additional infrastructure for the DLQ bucket
+ - Additional complexity with logic for checking blocks and merge-able events
```
leplatrem, line 66:
```suggestion
 - Additional complexity with logic for checking blocking and merge-able events 
```

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+- Overall performance of JBI: high
+- How intuitive the solution: low - users will notice data is missing but see status pages that look green
+
+**Pros:**
+- Allows for retries up to a designated amount of time
+- Keeping all services within the container make end-to-end testing and debugging easier (compared to option 4)
+
+**Cons:**
+- Increases complexity of the containers
+- Data will not persist container restarts
```
leplatrem, line 125:
```suggestion
- Data will not persist container restarts (ie. redeploy)
```

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+**Cons:**
+- Most complex solution
+- Highest effort for engineers to build and maintain
+
+### Option 6: A combination of the above
+Example: We could create a simple DLQ (a table in postgres) external queue for re-processing and then alert users if the DLQ grows too quickly.
+
+## Decision Outcome
+
+Pending discussion
```
leplatrem, line 176: Isn't this exposed in `## Proposed Solution`?

leplatrem (commented):
Some feedback after re-reading again the whole doc...

I haven't fully understood how you would distinguish from *blocking events* from others. Is it based on Jira HTTP responses? (ie. 401/500 vs. 400)

How do we merge events? What is the cardinality? `(product, bug #, event_type)` ?

I know that this ADR is about choosing an overall strategy and not going into implementation details, but I believe that the chosen solution would be easier to understand if some aspects were detailed.

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,185 @@
...
+- How intuitive the solution is to the users that depend on the data (will picking the easiest option solve their needs?)
+
+## Proposed Solution
+We propose to use a data bucket as DLQ. Events can attempt to be reprocessed from here, but we will skip any events identified as stale. See the diagram below:
+
+![Flow chart detailing the data flow, see expandable below for full details](./003.drawio.jpg "Proposed Solution Flow Chart")
+
+<details>
+
+  <summary>Braekdown of flowchart</summary>
```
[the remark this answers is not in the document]
alexcottner: Fixed this and several other typos. My first drafts are always rough for spelling.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Adding clarity and examples

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Adding clarity and examples

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Process ordering becomes important here, so I wouldn't want to delete a potentially blocking event before attempting to reprocess it just in case we receive another event around the same time.

Also, touching the same file with multiple operations quickly in a data bucket can be messy. GCP has a rate limit of 1 request per second per object. Sending a delete request followed by a write request shortly after might cause issues if we do it too quickly.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: Sorry, that's compared to Option 1 which is "log the failure and drop the data". In that scenario, users would be expected to manually correct any data that failed to sync unless they wanted to force a re-sync somehow from Bugzilla.

docs/adrs/003-bugzilla-response-codes.md
alexcottner: adjusting that, hopefully we'll finalize our approach tomorrow

docs/adrs/003-bugzilla-response-codes.md
```diff
@@ -0,0 +1,184 @@
+# Bugzilla disables webhooks after too many errors
+
+- Status: proposed
```
leplatrem, line 3: This is accepted now

leplatrem (approved):
I read again quickly, seems to match our implementation.

I think the format of this ADR is not exactly like others we have, where we list all options, and in the `Chosen solution` we only say which one we chose and why.

Examples:
- https://github.com/mozilla/remote-settings/blob/main/docs/adr/adr_002_packaging_releasing_admin_UI.md#decision-outcome
- https://github.com/mozilla-services/telescope/blob/main/docs/adrs/adr_0001_history.md#decision-outcome

But this may not be very important :)

PR #778, opened 2023-11-29
Inject `rid` into Jira Client error logs

grahamalama:
This PR adds a way for us to inject a request ID into errors logged in the Jira client. This will allow us to correlate erroring Jira calls with the rest of an action pipeline.

Considering this context manager along with our decorator, we're really fighting against the Jira client to make it custom to the point it's making more sense to roll our own. But this will hopefully make logs more actionable in the meantime.


jbi/services/jira.py
```diff
@@ -71,6 +72,19 @@ class JiraClient(Jira):
     in our instrumentation decorator.
     """
 
+    def __init__(self, *args, **kwargs):
+        self._rid = None
+        super().__init__(*args, **kwargs)
+
+    @contextlib.contextmanager
+    def rid(self, rid):
+        """Temporarily set rid to access during logging"""
```
leplatrem, line 81: maybe mention request ID here (instead of just `rid`)

jbi/services/jira.py
```diff
@@ -238,7 +257,8 @@ def get_issue(self, context: ActionContext, issue_key):
         """Return the Jira issue fields or `None` if not found."""
         logger.debug("Getting issue %s", issue_key, extra=context.model_dump())
         try:
-            response = self.client.get_issue(issue_key)
+            with self.client.rid(context.rid):
+                response = self.client.get_issue(issue_key)
```
leplatrem, line 261: I was wondering whether we could have something less verbose instead of duplicating the usage of the context manager...

```py
def client(self, rid):
    with self._client.rid(rid):
        return self._client
```

```py
response = self.client(rid).get_issue(issue_key)
```

leplatrem (approved):
💯

jbi/services/jira.py
grahamalama: I like that API too (`self.client(rid).<method>`), but I think returning the client from within the context manager will immediately trigger the `__exit__` method, right? That is:
```
client = self.client(123)
```

Will set `self._client._rid = 123`, return the client, then immediately set `self._client._rid = None` before we use it. I think we'll still need a bunch of `with` blocks wrapping actual method calls one way or another.

leplatrem: oh you're right 😵

leplatrem: only way would be to clone on each call to `self.client(rid)`

grahamalama:
Going to close this PR (for now) in favor of a logging filter approach to come

PR #783, opened 2023-12-06
Make fewer requests in heartbeat, handle exceptions

grahamalama:
This PR adjusts the heartbeat so that we make fewer requests in it.
 
Before, this PR we'd make `(n * 3) + 1` API calls, where `n` equals the number of configured projects:
  - `"all_projects_are_visible"`: - 1 API call
  - `"all_projects_have_permissions"`: `n` API calls (in a threadpool)
  - `"all_projects_components_exist"`: `n` API calls (not in a threadpool)
  - `"all_projects_issue_types_exist"`: `n` API calls (not in a threadpool)

With 22 configured prod projects, that's 67 API calls (😱).

Now, we make n + 3 API calls:
  - `"all_projects_are_visible"`: - 1 API call
  - `"all_projects_have_permissions"`: 1 API call
  - `"all_project_custom_components_exist"` (renamed): `n` API calls (in a threadpool)
  - `"all_projects_issue_types_exist"`: 1 API call

For a grand total of 25 API calls in prod.

This was done by using different API endpoints (or params on those endpoints) to fetch more results at once. It's annoying that Jira doesn't provide an API endpoint to fetch components by project, but we were luckily able to make an improvement

This PR also adds a little more exception handling to functions we call in the heartbeat


jbi/services/bugzilla.py
```diff
@@ -142,7 +142,10 @@ def check_health(self) -> ServiceHealth:
         all_webhooks_enabled = False
         if logged_in:
             jbi_webhooks = self.client.list_webhooks()
-            all_webhooks_enabled = len(jbi_webhooks) > 0
+            if len(jbi_webhooks) == 0:
+                logger.info("No webhooks enabled")
```
alexcottner, line 146: I'm a fan of returning as early as possible because then I know the rest of this function doesn't apply if these conditions are met. Does this read better to you if we return a health object immediately if `len(jbi_webhooks) == 0`? It would also remove the else statement.

jbi/services/jira.py
```diff
@@ -124,16 +136,24 @@ def check_health(self, actions: Actions) -> ServiceHealth:
...
         return health
 
     def _all_projects_visible(self, actions: Actions) -> bool:
-        visible_projects = {project["key"] for project in self.fetch_visible_projects()}
+        try:
+            visible_projects = {
+                project["key"] for project in self.fetch_visible_projects()
+            }
+        except requests.HTTPError:
+            logger.exception("Error fetching visible Jira projects")
```
alexcottner, line 154: Is it worth logging the exception thrown to be verbose?

alexcottner (approved):
Two small questions, but definitely good to approve.

jbi/services/jira.py
grahamalama: Python does this automagically when you do [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception), but good call

alexcottner: Well then, TIL

jbi/services/bugzilla.py
grahamalama: Added [1cf9906](https://github.com/mozilla/jira-bugzilla-integration/pull/783/commits/1cf9906c813d34a75e8f54aba29f263d596bf2e3) and [6b6098a](https://github.com/mozilla/jira-bugzilla-integration/pull/783/commits/6b6098acd978186add2876ab7d2e596768c4c1b1) (and tagged you for a re-review). Lmk what you think

alexcottner: Oh yeah that's really clean. I like that

grahamalama:
Tagged this issue as a "breaking change" since we renamed a field of the heartbeat payload

PR #786, opened 2023-12-07
Replace formatting tools and `pylint` with `ruff`

grahamalama:


bin/lint.sh
grahamalama, line 1: I honestly hate shell scripts and almost replaced this with a Python script with `argparse` in this PR. Could still be convinced, but I think this works for now.

alexcottner: Shell scripts (excluding powershell) don't bother me. But if this is annoying for most python devs to work with then my vote is to change it.

alexcottner (approved):

bin/lint.sh
leplatrem: As you prefer :)
Shell scripts are fine, as long as they are obvious and extremely simple. This one is on the edge..

Also, it's duplicated in the ctms-api repo, maybe we could write it once for good (in python?) and pull it from our syseng-pod repo?

PR #791, opened 2023-12-11
Change method for fetching visible projects and issue types in __heartbeat__

grahamalama:
From the commit message for fd461335341ddabae96714b2b0841023c4020a9d:

In v6.3.6, we bumped atlassian-python-api from 3.41.4. That bump included a [PR](https://github.com/atlassian-api/atlassian-python-api/pull/1270) which changed the way the library fetched projects. This somehow affected our project when we fetched visible projects and issue types in the heartbeat. We never fully determined the root cause, but the while loop in the linked PR was suspect.

In this commit, we make further improvements to the code that's involved in the heartbeat to make it more efficient.

Note that one of the tradeoffs we made for now was hardcoding the maximum number of projects that may be configured at 50. This is due to the way we fetch issue types by project. This is something that we can fix in the future, but is not a problem that we need to solve for now since we only have 26 projects configured.

### DIFF
--- a/jbi/services/jira.py
+++ b/jbi/services/jira.py
@@ -11,7 +11,7 @@ import concurrent
 import json
 import logging
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Iterable, Optional
+from typing import TYPE_CHECKING, Any, Collection, Iterable, Optional
 
 import requests
 from atlassian import Jira
@@ -94,7 +94,6 @@ class JiraClient(Jira):
 
     get_server_info = instrumented_method(Jira.get_server_info)
     get_project_components = instrumented_method(Jira.get_project_components)
-    projects = instrumented_method(Jira.projects)
     update_issue = instrumented_method(Jira.update_issue)
     update_issue_field = instrumented_method(Jira.update_issue_field)
     set_issue_status = instrumented_method(Jira.set_issue_status)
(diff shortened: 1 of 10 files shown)


jbi/services/jira.py
```diff
@@ -103,17 +102,55 @@ def raise_for_status(self, *args, **kwargs):
...
+
+        if not self.cloud:
+            raise ValueError(
+                "``projects_from_cloud`` method is only available for Jira Cloud platform"
+            )
+
+        params = {}
+
+        if keys is not None:
+            if len(keys) > 50:
```
alexcottner, line 127: Do we care to have this check in two places now? Is this trying to safeguard us in case we change where they array items come from in the future?

alexcottner (approved):

jbi/services/jira.py
grahamalama: > Is this trying to safeguard us in case we change where they array items come from in the future?

Yep, if/when we eventually use this function to return paginated results for more than 50 projects, we'll still want to ensure `keys < 50`. Say we had 100 configured projects -- this guard would be a reminder for us to use two requests with 50 keys each, since the API only allows 50 keys per request ([docs](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-projects/#api-rest-api-2-project-search-get)).

PR #792, opened 2023-12-12
Properly provide project keys to `Jira.paginated_projects`

grahamalama:
I assumed (wrongly) that this Jira package was [using requests](https://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls) to do param parsing, but it's actually using [urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) directly, which doesn't do the magic of turning `{"param": ["foo", "bar"]} --> ?param=foo&param=bar`

### DIFF
--- a/jbi/services/jira.py
+++ b/jbi/services/jira.py
@@ -121,17 +121,17 @@ class JiraClient(Jira):
                 "``projects_from_cloud`` method is only available for Jira Cloud platform"
             )
 
-        params = {}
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
I think this is a good example of why we want to allow feature branch deployments to a lower environment for testing.

PR #795, opened 2023-12-14
New structure for Jira and Bugzilla clients + services

grahamalama:
In this PR, we introduce a slightly new structure for the code we use to communicate with Jira and Bugzilla.

Before, we had a `services` package with a `jira` and `bugzilla` module. 
```
jbi
├── services
    ├── __init__.py
    ├── bugzilla.py
    ├── common.py
    └── jira.py
```
Each of those modules had a `*Client` and `*Service` class.

Now, we do essentially the same thing, except we have:

```
jbi
├── bugzilla
│   ├── __init__.py
│   ├── client.py
│   └── service.py
├── common
│   ├── __init__.py
│   └── instrument.py
└── jira
    ├── __init__.py
    ├── client.py
    └── service.py
```

this is to:
  - reinforce the separate concerns of a `client` and `service`, especially in tests
    - a client should only make HTTP requests and receive responses
    - services should use clients to do higher-level project operations
  - allow for the continued encapsulation of different domains. For example, instead of one `models` module, it might make sense to move certain models under specific namespaces

### DIFF
--- a/jbi/bugzilla/__init__.py
+++ b/jbi/bugzilla/__init__.py
@@ -1,0 +1,3 @@
+from .service import BugzillaService, get_service
+
+__all__ = ["get_service", "BugzillaService"]
--- a/jbi/bugzilla/__init__.py
+++ b/jbi/bugzilla/__init__.py
@@ -1,3 +1,2 @@
-from .service import BugzillaService, get_service
-
-__all__ = ["get_service", "BugzillaService"]
+from .service import BugzillaService as BugzillaService
+from .service import get_service as get_service
--- a/jbi/jira/__init__.py
+++ b/jbi/jira/__init__.py
@@ -1,3 +1,2 @@
-from .service import JiraService, get_service
-
-__all__ = ["JiraService", "get_service"]
+from .service import JiraService as JiraService
+from .service import get_service as get_service
--- a/jbi/jira/__init__.py
+++ b/jbi/jira/__init__.py
@@ -1,0 +1,3 @@
+from .service import JiraService, get_service
+
+__all__ = ["JiraService", "get_service"]
--- a/jbi/bugzilla/client.py
+++ b/jbi/bugzilla/client.py
@@ -1,0 +1,126 @@
+import logging
+
+import requests
+
+from jbi import environment
+from jbi.common.instrument import instrument
+from jbi.models import (
+    BugzillaApiResponse,
+    BugzillaBug,
+    BugzillaComment,
+    BugzillaComments,
+    BugzillaWebhooksResponse,
+)
+
+settings = environment.get_settings()
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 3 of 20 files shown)


jbi/jira/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .service import JiraService, get_service
+
+__all__ = ["JiraService", "get_service"]
```
leplatrem, line 3: nit: same here, they seem to only be used in tests, which could import from `jbi.jira.service`

grahamalama: I was thinking about this too. Without putting these in `__all__`, Ruff was failing with [F401](https://docs.astral.sh/ruff/rules/unused-import/). Adding `__all__` "fixed" it, but reexporting ([17dae59](https://github.com/mozilla/jira-bugzilla-integration/pull/795/commits/17dae594eee4fc08d36a603162692482925d12f3)) makes the intent more explicit. 

<details><summary>We could also do something like this</summary>
<p>

```diff
diff --git a/jbi/bugzilla/__init__.py b/jbi/bugzilla/__init__.py
index e680c38..4bc0a4e 100644
--- a/jbi/bugzilla/__init__.py
+++ b/jbi/bugzilla/__init__.py
@@ -1,2 +1 @@
-from .service import BugzillaService as BugzillaService
-from .service import get_service as get_service
+from .service import BugzillaService, get_service
diff --git a/jbi/jira/__init__.py b/jbi/jira/__init__.py
index d9af23f..d9964cd 100644
--- a/jbi/jira/__init__.py
+++ b/jbi/jira/__init__.py
@@ -1,2 +1 @@
-from .service import JiraService as JiraService
-from .service import get_service as get_service
+from .service import JiraService, get_service
diff --git a/pyproject.toml b/pyproject.toml
index bdd45f5..02b9ae6 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,9 @@ build-backend = "poetry.core.masonry.api"
 target-version = "py312"
 extend-select = ["I"]
 
+[tool.ruff.lint.per-file-ignores]
+"__init__.py" = ["F401"] # useful for reexporting symbols 
+
 [tool.pytest.ini_options]
 testpaths = [
     "tests/unit",
```

</p>
</details> 


Though I think reexporting doesn't hurt for now.

grahamalama: Also, as far as where they're used:

`get_service` is only used in the runner, but the `*Service` classes are used in more places, usually for typing purposes

jbi/bugzilla/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .service import BugzillaService, get_service
+
+__all__ = ["get_service", "BugzillaService"]
```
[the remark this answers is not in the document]
grahamalama: Replied to the other, similar comment 👍🏻

PR #796, opened 2023-12-18
Mark all pydantic models as `frozen`

grahamalama:
This PR sets [`frozen=True`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.frozen) on our Pydantic models to enforce "immutability". As immutable as you can get in Python, anyway. 

This change actually didn't have much of an impact on our application code, which is good. Most of the changes involved our test code. These included:

- instead of using a model instance as a fixture then modifying the data in the tests, use factories to create a model with the desired attributes directly in the tests
- in instances where we used a factory without overriding any kwargs (e.g. `bugzilla_webhook_factory()`), use the fixture [provided by pytest-factoryboy when registering the factory](https://pytest-factoryboy.readthedocs.io/en/stable/#model-fixture) instead

### DIFF
--- a/jbi/bugzilla/service.py
+++ b/jbi/bugzilla/service.py
@@ -93,11 +93,11 @@ class BugzillaService:
     def refresh_bug_data(self, bug: BugzillaBug):
         """Re-fetch a bug to ensure we have the most up-to-date data"""
 
-        updated_bug = self.client.get_bug(bug.id)
+        refreshed_bug_data = self.client.get_bug(bug.id)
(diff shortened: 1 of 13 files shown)


alexcottner (approved):
I dislike something about having a factory specifically for unit tests. Feels equivalent to unsprung weight for some reason. But I'm not sure what reasonable alternatives we'd have in this situation.

leplatrem (approved):

PR #803, opened 2024-01-10
Relocate Bugzilla models to `bugzilla` namespace

grahamalama:
This PR continues the work of https://github.com/mozilla/jira-bugzilla-integration/pull/795 to reduce coupling between the Jira and Bugzilla domains. 

Here, we move all Bugzilla-specific models under the `bugzilla` name space. The intention is that this `bugzilla.models` module be used either internally by the Bugzilla Service / Client or by JBI code where both the Jira and Bugzilla services are used together (like in `steps`). The Bugzilla models should not be needed in the Jira domain and vice versa, except in the _rare_ occasion of maybe type checking.


alexcottner (approved):

jbi/runner.py
```diff
@@ -175,7 +194,7 @@ def execute_action(
 
         runner_context = runner_context.update(bug=bug)
         try:
-            action = bug.lookup_action(actions)
```
leplatrem, line 178: 👌

leplatrem (approved):

PR #836, opened 2024-02-01
Fix #723: format comment with Jira syntax

### LINKED ISSUE #723: Bugzilla comments with Markdown are misformatted in Jira
Hi! I recently left [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1858542#c6) on a Bugzilla bug, with some Markdown formatting: links, lists, headings, and code blocks. It was imported into Jira [like this](https://mozilla-hub.atlassian.net/browse/DISCO-2633?focusedCommentId=785864). (Please let me know if the Jira comment is inaccessible; I can attach a screenshot!)

* The numbered and ordered lists were formatted correctly.
* The links lost their formatting, and look like raw Markdown in Jira.
* The heading (`###`) looks like it was formatted as a numbered list in Jira.
* The triple-backtick code block was also imported as raw Markdown.

leplatrem:
This is best effort, but should cover most cases.

A proper implementation would probably require using a pyparsing like https://pypi.org/project/jira2markdown/ does.

### DIFF
--- a/jbi/jira/md2jira.py
+++ b/jbi/jira/md2jira.py
@@ -74,12 +74,14 @@ def convert_line(line: str) -> str:
     line = re.sub(r"\[(.+?)\]\((.+?)\)", r"[\1|\2]", line)
     # Strikethrough
     line = re.sub(r"~~(.+?)~~", r"-\1-", line)
+    # Italic.
+    # _this_ but not __this__
+    line = re.sub(r"([^_])_([^_]+?)_", r"\1_\2_", line)
+    # *this* but not **this**
+    line = re.sub(r"([^\*])\*([^\*]+?)\*", r"\1_\2_", line)
     # Bold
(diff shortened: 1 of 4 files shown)


grahamalama:
> A proper implementation would probably require using a pyparsing

It might be heavyweight, but we could also pull in Pandoc for our conversion needs -- either by using subprocess or a [Python wrapper](https://pypi.org/project/pandoc/)

leplatrem:
> It might be heavyweight, but we could also pull in Pandoc for our conversion needs -- either by using subprocess or a [Python wrapper](https://pypi.org/project/pandoc/)

I don't see that it would transform to Jira formatting syntax.

Anyway, this regexp-based implementation may have limitations, but it is far better than posting raw markdown and was cheap to implement. It should cover most cases (see tests)

Issue #234, opened 2022-09-06
Bugzilla Data Modeling (Optional Fields)


bsieber-mozilla:
Current Models in JBI differ from actual models.

Since the Webhook was designed with different models than the API, the models used in JBI should also represent this reality.

i.e.
Webhook Comments =/= API Comments
Webhook Bugs =/= API Bugs

PR #839, opened 2024-02-02
Fix #723: turn markdown into Jira syntax on comments and description

### LINKED ISSUE #723: Bugzilla comments with Markdown are misformatted in Jira

leplatrem:
Fix #723

Use `pandoc` instead of #836

### DIFF
--- a/bin/healthcheck.py
+++ b/bin/healthcheck.py
@@ -12,5 +12,9 @@ session = requests.Session()
 retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
 session.mount("http://", HTTPAdapter(max_retries=retries))
 if __name__ == "__main__":
-    response = session.get(f"http://0.0.0.0:{PORT}/")
+    url = f"http://0.0.0.0:{PORT}"
+    response = session.get(f"{url}/")
     response.raise_for_status()
+
+    hb_response = session.get(f"{url}/__heartbeat__")
+    assert hb_response.json["pandoc_install"]
--- a/bin/healthcheck.py
+++ b/bin/healthcheck.py
@@ -2,19 +2,29 @@
 
 import os
 
+import backoff
 import requests
-from requests.adapters import HTTPAdapter, Retry
 
-PORT = os.environ["PORT"]
+PORT = os.getenv("PORT", "8000")
 
-session = requests.Session()
 
-retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
-session.mount("http://", HTTPAdapter(max_retries=retries))
-if __name__ == "__main__":
+@backoff.on_exception(
+    backoff.expo,
+    requests.exceptions.RequestException,
+    max_tries=5,
+)
+def check_server():
     url = f"http://0.0.0.0:{PORT}"
-    response = session.get(f"{url}/")
+    response = requests.get(f"{url}/")
     response.raise_for_status()
 
-    hb_response = session.get(f"{url}/__heartbeat__")
-    assert hb_response.json["pandoc_install"]
+    hb_response = requests.get(f"{url}/__heartbeat__")
+    hb_details = hb_response.json()
+    # Check that pandoc is installed, but ignore other checks
+    # like connection to Jira or Bugzilla.
+    assert hb_details["jira"]["pandoc_install"]
+    print("Ok")
+
+
+if __name__ == "__main__":
+    check_server()
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -215,7 +215,9 @@ class JiraService:
         fields: dict[str, Any] = {
             "summary": bug.summary,
             "issuetype": {"name": issue_type},
-            "description": markdown_to_jira(description[:JIRA_DESCRIPTION_CHAR_LIMIT]),
+            "description": markdown_to_jira(
+                description, max_length=JIRA_DESCRIPTION_CHAR_LIMIT
+            ),
             "project": {"key": context.jira.project},
         }
         logger.debug(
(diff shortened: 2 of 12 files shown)


bin/healthcheck.py
```diff
@@ -12,5 +12,9 @@
 retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
 session.mount("http://", HTTPAdapter(max_retries=retries))
 if __name__ == "__main__":
-    response = session.get(f"http://0.0.0.0:{PORT}/")
+    url = f"http://0.0.0.0:{PORT}"
+    response = session.get(f"{url}/")
     response.raise_for_status()
+
+    hb_response = session.get(f"{url}/__heartbeat__")
```
leplatrem, line 19: this shouldn't be retried, because we don't `raise_for_status()`. 

Two options:
- revert 86a3c3b8aa0816
- rewrite this healthcheck using `@backoff` so that it retries only when exceptions are raised

grahamalama:
I'm guessing we need to `sudo apt-get install -y pandoc` Somewhere in the `run_test` job 👀

jbi/jira/service.py
```diff
@@ -213,7 +215,7 @@ def create_jira_issue(
         fields: dict[str, Any] = {
             "summary": bug.summary,
             "issuetype": {"name": issue_type},
-            "description": description[:JIRA_DESCRIPTION_CHAR_LIMIT],
+            "description": markdown_to_jira(description[:JIRA_DESCRIPTION_CHAR_LIMIT]),
```
grahamalama, line 218: I wonder what happens if we cut off a comment in the middle of some markup, like `[some link](https://www.ex`, and then that's the character limit. Will `pandoc` handle it? Throw some exception about malformed input? I wonder if we need a smarter truncation of the comment so we break on the last complete word before the character limit 🤔

jbi/jira/service.py
```diff
@@ -257,7 +259,7 @@ def add_jira_comment(self, context: ActionContext):
 
         issue_key = context.jira.issue
         formatted_comment = (
-            f"*({commenter})* commented: \n{{quote}}{comment.body}{{quote}}"
+            f"*{commenter}* commented: \n{markdown_to_jira(comment.body or "")}"
```
grahamalama, line 262: Not something to address in this PR, but it's slightly annoying to have to account for a `comment.body` being `None`. I wonder if that's something that's even possible from the Bugzilla side. Like I'd almost consider a comment without a body a malformed payload and something we should refuse (but still return a `200` for webhook `ack` purposes)

grahamalama (commented):
Looks good! The only thing I'd like to check on before we merge this is my question about handling the character limit gracefully. After that, g2g

jbi/jira/service.py
leplatrem: you are right, there is a broader question here, how does pandoc handle malformed input! 

truncating at the last complete word is a good idea, I'll have a look what I can do.

jbi/jira/service.py
leplatrem: Yes
And I think this relates to https://github.com/mozilla/jira-bugzilla-integration/issues/234

tests/unit/jira/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+    this was -wrong-.
+    """
+    ).strip()
+
+    assert markdown_to_jira(markdown) == jira
+
+
+def test_markdown_to_jira_with_malformed_input():
+    assert markdown_to_jira("[link|http://noend") == "\\[link|http://noend"
```
grahamalama, line 53: Nice that Pandoc handles this no problem, but wish it'd return an empty string rather than the mangled output `\\[link|http://noend`. But overall, we're in a much better place than when we started. Maybe worth spinning an issue up for this for polish, but good enough for this PR.

grahamalama (approved):

tests/unit/jira/test_utils.py
leplatrem: I'm not sure about your suggestion, in markup languages when you make a syntax mistake, it shows the markup "as is" [example[link] or [example](link

grahamalama: Hmm I guess you're right. This is what I'd want though:

```
>>> markdown_to_jira("The line is [too long|https://www.example.com]", max_len=20)
"This line is ..."
```

I think this would rely on us converting the entire comment to the new markup, then looking for a good place to truncate after the fact. 

I don't know if this sort of thing even makes sense, but in my head, Pandoc would consider `[too long|https://www.example.com]` as an entire "block" or "unit of markup" (as it would `**bold**` or `## Heading`), and then split on a block, rather than on a space in the middle of a "block".

leplatrem: Ok, I understand what you mean now.

Well, we split on space so the result would be:

```
>>> markdown_to_jira("The line is [too long|https://www.example.com]", max_len=20)
"This line is [too"
```

We could also truncate on `max_length - 3` and add `...`  or something more explicit like `max_length - len("[truncated...]")`


> I think this would rely on us converting the entire comment to the new markup, then looking for a good place to truncate after the fact.

Well, it would require having access to the semantics of characters in their context, not just the output of the new markup

PR #860, opened 2024-02-14
Fix #857: Support Basic Auth for the JBI Api Key

### LINKED ISSUE #857: Add support for basic auth in protected endpoints
With basic auth support, users visiting `/powered_by_jbi/` would be prompted a username/password, and could access the content providing the JBI APi key as the password for example (with blank username)

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

leplatrem:
Fix #857

### DIFF
--- a/jbi/router.py
+++ b/jbi/router.py
@@ -8,7 +8,7 @@ from typing import Annotated, Optional
 from fastapi import APIRouter, Body, Depends, HTTPException, Request, Response, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import HTMLResponse
-from fastapi.security import APIKeyHeader
+from fastapi.security import APIKeyHeader, HTTPBasic, HTTPBasicCredentials
 from fastapi.templating import Jinja2Templates
 
 from jbi import bugzilla, jira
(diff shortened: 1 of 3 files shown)


jbi/router.py
```diff
@@ -75,15 +75,22 @@ def version(version_json: VersionDep):
...
 ):
-    if not secrets.compare_digest(api_key, settings.jbi_api_key):
+    if not api_key and basic_auth:
+        api_key = basic_auth.password
+    if not api_key or not secrets.compare_digest(api_key, settings.jbi_api_key):
         raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED, detail="Forbidden"
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Incorrect API Key",
+            headers={"WWW-Authenticate": "Basic"},
```
grahamalama, line 93: Nice catch. Along with the "Basic" challenge in the response, is there a good name for what we do with the `X-Api-Key` header?

grahamalama (approved):

jbi/router.py
leplatrem: If we would read it from the `Authorization` header, we could add various `WWW-Authenticate` values, but as-is there's no way to declare that auth type to the client

PR #861, opened 2024-02-14
Fix #833: Wait for auto-labeler for 10sec on new PRs

### LINKED ISSUE #833: Labeler action fights with `Force pull-requests label(s)` action
We have [an action](https://github.com/mozilla/jira-bugzilla-integration/blob/main/.github/workflows/labels.yaml) that verifies that PRs include labels.

We have [another action](https://github.com/mozilla/jira-bugzilla-integration/blob/main/.github/workflows/labeler..yaml) that automatically adds labels to certain issues.

The former action doesn't seem to accept labels that are applied by the latter. It'd be nice to fix this.

leplatrem:
Fix #833

### DIFF
--- a/.github/workflows/labels.yaml
+++ b/.github/workflows/labels.yaml
@@ -4,7 +4,16 @@ on:
   pull_request:
     types: [opened, labeled, unlabeled]
 jobs:
+  wait-triage:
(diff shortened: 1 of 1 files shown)


alexcottner (approved):
😎

PR #862, opened 2024-02-15
Fix #702: heartbeat now returns 200 on warnings

### LINKED ISSUE #702: Distinguish warnings from failures in heartbeat view
Currently, the heartbeat checks various things:
- makes sure Bugzilla credentials are correct
- makes sure all Bugzilla webhooks are enabled
- logs warning when number of errors in above threshold
- makes sure Jira credentials are correct
- makes sure all Jira projects are visible with the configured credentials
- same for Jira components and issues types

There has been many conversations about the role of the heartbeat endpoint.
When heartbeat fails, SRE need to take action (it's an incident).

One possible change could be:
- fail (ie. return 503) when something that would break all workflows
- warn (ie. return 200 and log warning) when only certain workflows would be affected

The main important goal is:
- we should not discover issues via reports of our users
- SRE should trust our heartbeat (no false negative)
- Jira admins should be warned when something on the Jira side has to be fixed

See also:
- https://github.com/mozilla/jira-bugzilla-integration/issues/645
- #21
- #390 
- #343
- [Notes about JBI monitoring](https://docs.google.com/document/d/1jjkBCAUP4kTTnp2WR6Dz8A_XiAoU4x9AFZbN6rfQEKs/edit#heading=h.yokk4gkginf9)

leplatrem:
Fix #702

### What's changed:

- leverage `dockerflow` check system (body of heartbeat response has changed)
- return 200 when the status is warning only
- return warning from checks when only a portion of workflows are affected
- let dockerflow log messages and details

### Questions:

- A lot of testing code in `test_router.py` asserts the content of the heartbeat response body. Shall we only leave the bare minimum (eg. all success, one warning, one error) instead of testing all combinations?  
- This pattern to do injection in checks is not great:
```py
def check_jira_all_project_issue_types_exist(service=None, actions=None):
    actions = actions or get_actions()
    service = service or get_service()
    ...
```
What would you recommend instead?
(*Note: `configuration.get_actions()` is not cached*)

### Todo Next:

- Wait for https://github.com/mozilla-services/python-dockerflow/pull/81 to leverage FastAPI heartbeat view

### DIFF
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -11,10 +11,11 @@ from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Iterable, Optional
 
 import requests
+from dockerflow import checks
 from requests import exceptions as requests_exceptions
 
 from jbi import Operation, bugzilla, environment
-from jbi.common.instrument import ServiceHealth
+from jbi.configuration import get_actions
 from jbi.jira.utils import markdown_to_jira
 from jbi.models import ActionContext
 
@@ -22,7 +23,7 @@ from .client import JiraClient, JiraCreateError
 
 # https://docs.python.org/3.11/library/typing.html#typing.TYPE_CHECKING
 if TYPE_CHECKING:
-    from jbi.models import Actions
+    pass
 
 settings = environment.get_settings()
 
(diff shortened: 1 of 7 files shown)


jbi/jira/service.py
```diff
@@ -520,3 +387,187 @@ def get_service():
...
+    service = service or get_service()
+    actions = actions or get_actions()
+
+    # Do not bother executing the rest of checks if connection fails.
+    if messages := check_jira_connection():
+        return messages
+
+    results = []
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
```
alexcottner, line 477: Hard coding this to 4 feels weird. Can we do some math based on CPU threads?

pyproject.toml
```diff
@@ -70,7 +70,7 @@ warn_return_any = true
 plugins = "pydantic.mypy"
 
 [[tool.mypy.overrides]]
-module = ["ruamel", "bugzilla", "atlassian", "atlassian.rest_client", "statsd.defaults.env"]
+module = ["ruamel", "bugzilla", "atlassian", "atlassian.rest_client", "statsd.defaults.env", "dockerflow"]
```
alexcottner, line 73: Are we missing a poetry.lock change?

leplatrem: I don't think `mypy` config ends up in the lock file

grahamalama (commented):
> This pattern to do injection in checks is not great:
```
def check_jira_all_project_issue_types_exist(service=None, actions=None):
    actions = actions or get_actions()
    service = service or get_service()
```
> What would you recommend instead?


I don't know how much better or worse this is, but here are some ideas:


- for providing actions, we can use the [global constant](https://github.com/mozilla/jira-bugzilla-integration/blob/a5a7fc64a809002fd063e46345abccdb5a4bc343/jbi/configuration.py#L37) we define on app initialization
- for providing services, maybe we could use [`register_partial`](https://github.com/mozilla-services/python-dockerflow/blob/2ba025470867433ebea93bd0911ddbc799beaeaf/src/dockerflow/checks/registry.py#L65)?

So that'd be:
```python
def check_jira_all_project_issue_types_exist(service, actions=ACTIONS):
    ...

checks.register_partial(check_jira_all_project_issue_types_exist, get_service())
```

but also, maybe these functions actually deserve to be service methods if we continually need to provide `service` as the first argument (though I don't know how compatible that is with the check registration system)

```python
@checks.register(name="jira.all_project_issue_types_exist")
def check_jira_all_project_issue_types_exist(self, actions=ACTIONS):
    ...
```

jbi/jira/service.py
```diff
@@ -520,3 +390,187 @@ def get_service():
...
+    service = service or get_service()
+    actions = actions or get_actions()
+
+    # Do not bother executing the rest of checks if connection fails.
+    if messages := check_jira_connection():
+        return messages
+
+    results = []
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=CPU_COUNT) as executor:
```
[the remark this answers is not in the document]
alexcottner: That would be fine with me, assuming we/forks aren't restricting which threads the application can run on. Probably a safe assumption.

jbi/router.py
```diff
@@ -45,21 +46,26 @@ def root(request: Request, settings: SettingsDep):
...
-    for health in health_map.values():
-        health_checks.extend(health.values())
-    if not all(health_checks):
-        response.status_code = 503
-    return health_map
+
+    if check_results.level < dockerflow_checks.ERROR:
+        response.status_code = 200
+    else:
+        response.status_code = 500
```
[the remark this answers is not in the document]
leplatrem: What do you think of https://github.com/mozilla-services/python-dockerflow/pull/98 ?

grahamalama:
> The problem with register_partial is that the parameter evaluated at registration time, and our mock_jira and mock_bugzilla come too late to override them (in test_router.py for example)

Ahh okay 👍🏻 

> but I'm not 100% convinced

Yeah, I'm not either. Nothing feels like the "right" solution here. Let's merge this with what feels "most right", then after https://github.com/mozilla/jira-bugzilla-integration/pull/874, we can consider how to make further improvements.

I started to mess around with this branch, which I pushed up [here](https://github.com/mozilla/jira-bugzilla-integration/tree/702-WIP) if you're interested ([and comparing the two](https://github.com/mozilla/jira-bugzilla-integration/compare/702-dockerflow-heartbeat..702-WIP)).

So far, it ended up being more steps to the same problem of 

> parameter evaluated at registration time, and our mock_jira and mock_bugzilla come too late to override them

I think for that branch to work, we'd need to make a fixture that actually generates a new app for each test (by calling `FastAPI(...)`)

So make that'd mean we make our app with a factory pattern. Like in `jbi/app.py`:

```python

def make_app(*args, **kwargs):
    ...
    return FastAPI(...)


app = make_app(*args, **kwargs)
```

and in conftest:

```python
def app:
    return make_app()
```

that way we could ensure the the client is patched before we register the checks.

But again, let's get this and #874 merged before we go down this road. We might even look to Python Dockerflow to figure out how to best handle "provide dependencies to checks that need them".

PR #884, opened 2024-02-27
Fix #645: Switch log messages to INFO level

### LINKED ISSUE #645: Assess the state of our logging

leplatrem:
Fix #645

Debug level seemed like a good idea, because in the end, that's why we need those log messages, to debug stuff. 

But debug messages are not enabled in default deployment config, and do not seem to be collected in our Logging panel.

We could ask ops to reconfigure it the service/logging (for the second time), but I believe that using info is a reasonable fix to this problem.

The trafic is low enough, these messages are globally very concise compared to ridiculous amount of logs coming from k8s

### DIFF
--- a/tests/unit/jira/test_service.py
+++ b/tests/unit/jira/test_service.py
@@ -162,7 +162,6 @@ def test_get_issue_handles_404(
     assert return_val is None
 
     before, after = capturelogs.records
-    assert before.levelno == logging.DEBUG
     assert before.message == "Getting issue JBI-234"
 
     assert after.levelno == logging.ERROR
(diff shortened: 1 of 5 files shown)


tests/unit/jira/test_service.py
```diff
@@ -180,7 +179,6 @@ def test_get_issue_reraises_other_erroring_status_codes(
             jira_service.get_issue(context=action_context, issue_key="JBI-234")
 
     [record] = capturelogs.records
-    assert record.levelno == logging.DEBUG
```
alexcottner, line 183: Shouldn't we switch these to `logging.INFO` rather than deleting them?

leplatrem: Hmm, why not, although this is not really something we assert usually

alexcottner (approved):

Issue #464, opened 2023-04-06
Provide ability to link Bugzilla `severity` field to a Jira field


grahamalama:

leplatrem:
Could it be a `maybe_update_severity` step with parameters like?

```
- whiteboard_tag: sp3
  bugzilla_user_id: 396948
  description: Speedometer 3
  parameters:
    jira_severity_field: severity
    jira_severity_map:
      S1: Very bad
      S2: Bad
      S3: Important
      S4: Not important
      N/A: Unknown
```

leplatrem:
Related #247

leplatrem:
This was done in #889

PR #889, opened 2024-03-01
Fix #247 #464: add `maybe_update_issue_priority` and `maybe_update_issue_severity` steps

### LINKED ISSUE #247: Sync priority and severity fields from bugzilla to JIRA
As per summary. I wrote support for this in the older rust comparison tool that Blake wrote, and my mapping looks like this for priorities:

```
            "P1" => "1".to_string(),
            "P2" => "2".to_string(),
            "P3" => "3".to_string(),
            "P4" => "4".to_string(),
            "P5" => "5".to_string(),
            "--" => "10000".to_string(),
```

There are other values in JIRA but the 1-5 values appear builtin (and are visually represented as highest through to lowest, respectively) and directly match P1-P5 so that part seems straightforward.

Severity fields, at least in the FIDEFE project, are equivalent to bugzilla visually in the UI - I haven't investigated if the underlying values are strings with the same value or if, like priorities, there is some kind of custom internal value identifier for them.

leplatrem:
Fix #247 #464 

* [x] Basic code
* [x] Test manually against real server

See https://developer.atlassian.com/server/jira/platform/jira-rest-api-examples/#creating-an-issue-using-custom-fields

### Priority

The `priority` is not custom, and take the text value coming from the mapping as a `name` subfield:

```python
>>> s.client.update_issue_field(key="SE-3850", fields={"priority": {"name": "High"}})
>>>
```
Screenshot 2024-03-05 at 17 12 02

### Severity

There are two severity fields (one with trailing space 🙄), but both are defined with the same type and take the mapped value as a `value` subfield:

```python
>>> [f for f in s.client.get_all_fields() if "severity" in f["name"].lower()]
[
{'id': 'customfield_10716', 'key': 'customfield_10716', 'name': 'Severity', 'untranslatedName': 'Severity', 'custom': True, 'orderable': True, 'navigable': True, 'searchable': True, 'clauseNames': ['cf[10716]', 'Severity', 'Severity[Dropdown]'], 'schema': {'type': 'option', 'custom': 'com.atlassian.jira.plugin.system.customfieldtypes:select', 'customId': 10716}}, 

{'id': 'customfield_10319', 'key': 'customfield_10319', 'name': 'Severity ', 'untranslatedName': 'Severity ', 'custom': True, 'orderable': True, 'navigable': True, 'searchable': True, 'clauseNames': ['cf[10319]', 'Severity ', 'Severity [Dropdown]'], 'schema': {'type': 'option', 'custom': 'com.atlassian.jira.plugin.system.customfieldtypes:select', 'customId': 10319}}
>>>
```
```

```python
>>> s.client.update_issue_field(key="FIDEFE-4686", fields={"customfield_10319": {"value": "S2"}})
>>>
```

Note, that in order to synchronize this severity field, the following parameter will have to be put in the config file:
```yaml
  parameters:
    jira_project_key: FIDEFE
    jira_severity_field = customfield_10319
```
because navigating the API to guess the proper severity field depending on the configured update screen seems fragile and complicated.

### DIFF
--- a/tests/unit/test_steps.py
+++ b/tests/unit/test_steps.py
@@ -757,7 +757,7 @@ def test_update_issue_priority(
     )
 
     mocked_jira.create_issue.assert_not_called()
-    mocked_jira.update_issue_field(
+    mocked_jira.update_issue_field.assert_called_with(
         key="JBI-234", fields={"priority": {"name": "Urgent"}}
     )
 
@@ -782,7 +782,7 @@ def test_update_issue_severity(
 
     params = action_params_factory(
         jira_project_key=action_context.jira.project,
-        status_map={
+        severity_map={
             "S3": "Moderate",
         },
     )
@@ -791,8 +791,8 @@ def test_update_issue_severity(
     )
 
     mocked_jira.create_issue.assert_not_called()
-    mocked_jira.update_issue_field(
-        key="JBI-234", fields={"severity": {"name": "Moderate"}}
+    mocked_jira.update_issue_field.assert_called_with(
+        key="JBI-234", fields={"customfield_10716": {"value": "Moderate"}}
     )
 
 
--- a/tests/unit/test_steps.py
+++ b/tests/unit/test_steps.py
@@ -728,6 +728,74 @@ def test_change_to_unknown_resolution_with_resolution_map(
     ]
 
 
+def test_update_issue_priority(
+    action_context_factory,
+    mocked_jira,
+    action_params_factory,
+    webhook_event_change_factory,
+):
+    action_context = action_context_factory(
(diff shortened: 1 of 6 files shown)


tests/unit/test_steps.py
```diff
@@ -728,6 +728,74 @@ def test_change_to_unknown_resolution_with_resolution_map(
     ]
 
 
+def test_update_issue_priority(
```
alexcottner, line 731: I'd like to see some tests an invalid priority and an invalid severity being provided. Just so we don't accidentally make a change in the future.

alexcottner (approved):
Approving since I'm only asking for a couple more test scenarios and we can merge after that.

PR #903, opened 2024-03-08
[draft] Dead letter queue backend

leplatrem:
* [ ] Improve parametrized tests
* [ ] Add monitoring
* [ ] Add logging
* [ ] Decide whether it's acceptable to look-up the queue before checking whether there is an associated project in configuration (because it saves us a call to bugzilla to refresh the bug info)
* [ ] Find a design where we can increment a number of retries
* [ ] Add high-level methods to `QueueItem` for stuff like `age`, `repr()`, etc.


jbi/queue.py
```diff
@@ -0,0 +1,207 @@
...
+        # Sort by event datetime ascending.
+        sorted_existing = [
+            sorted(v, key=lambda i: i.payload.event.time or 0)
+            for v in existing.values()
+            if len(v) > 0
+        ]
+        by_oldest_bug_events = sorted(
+            sorted_existing, key=lambda items: items[0].payload.event.time or 0
+        )
+        return list(itertools.chain(*by_oldest_bug_events))
```
leplatrem, line 201: After thinking about this, we shouldn't flatten this list. But if the scheduler fails to process one request, it will want to jump to the next bug. 

```python

for items in all_items:
    for item in items:
        try:
            execute_action(item.payload, actions)
        except IgnoreInvalidRequestError:
            await queue.done(item)
        except Exception as exc:
            # Jump to next bug
            break
        else:
            await queue.done(item)
```

jbi/queue.py
```diff
@@ -0,0 +1,207 @@
...
+class PythonException(BaseModel, frozen=True):
+    type: str
+    description: str
+    details: str
+
+    @classmethod
+    def from_exc(cls, exc: Exception):
+        return PythonException(
+            type=exc.__class__.__name__,
+            description=str(exc),
```
leplatrem, line 32: For easier debugging I think we could dump the response status and body for `HTTPError`, with a `isinstance(exc, HTTPError) and hasattr(exc, "response")`

jbi/queue.py
```diff
@@ -73,8 +73,9 @@ async def get(self, bug_id: int) -> list[QueueItem]:
     async def get_all(self) -> dict[int, list[QueueItem]]:
         pass
 
+    @property
     @abstractmethod
-    async def size(self) -> int:
+    def size(self) -> int:
```
leplatrem, line 78: Are you sure about this one?
I was thinking that if we end up implementing the queue in a db/redis or whatever, we would need it to be async... 
Maybe YAGNI, just checking :)

grahamalama:
Going to close this in favor of https://github.com/mozilla/jira-bugzilla-integration/pull/918 since the scope of the PR has changed. First we're going to only add the queue class and an integration with the heartbeat, then we'll follow on with fully integrating it into the app

PR #912, opened 2024-03-12
Bug 1884727 - Make omc config update issue points

dmose:
The OMC team would like to add the syncing of story points without changing anything else.  Our [current configuration](https://github.com/mozilla/jira-bugzilla-integration/blob/07cb9b1fad7d82bacb7f14884b08921d4b360794/config/config.prod.yaml#L261-L271) doesn't have a `steps` stanza, and so is presumably just using the defaults (is that correct?).

In any case, I couldn't see a way to say "use the defaults except with these two changes", which was the semantics I really wanted, so I cloned the defaults from the code itself, which (I think) should get me almost the same thing, except that it doesn't continue to get new defaults as steps get added to the defaults over time.  Correct?

Assuming those things are correct, I'd love to get this reviewed...

CC @aminomancer

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -263,6 +263,21 @@
   description: Firefox Messaging System whiteboard tag
   parameters:
     jira_project_key: OMC
+    steps:
+      new:
+        - create_issue
+        - maybe_delete_duplicate
+        - add_link_to_bugzilla
+        - add_link_to_jira
+        - sync_whiteboard_labels
+        - maybe_update_issue_points
+      existing:
+        - update_issue_summary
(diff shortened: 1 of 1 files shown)


alexcottner (approved):

dmose:
I don't seem to have the privs to merge this PR, even though it's been approved. What needs to happen next?

leplatrem:
I just tagged a version with the commits of this PR, the changes should be rolled out to stage/prod soon. Check out the `#jbi` channel for updates 🙏

dmose:
Thanks; it seems to be working!

Issue #567, opened 2023-06-09
The lack of explicit connection between webhooks and workflows makes webhook errors hard to diagnose


grahamalama:
In Bugzilla, up to this point, we've created a new webhook per product. We have a naming convention of `JBI-<environment>-<product>`. For example:

| Name | Product | Component |
|--------|--------|--------|
| JBI-PROD-Core | Core | ANY |
| JBI-PROD-DataPlatformAndTools | Data Platform and Tools | ANY |
| JBI-PROD-DevTools | DevTools | ANY |

In config, projects specify the Jira project they map to and what whiteboard tag they use. There is no information indicating which product or products a workflow uses. 

This makes webhook errors hard to diagnose. For example, if I were to log in and see that the `JBI-PROD-Core` webhook had errors piling up and was disabled, I wouldn't have the information to know which project likely caused that blockage at a glance.

leplatrem:
Should we add the webhook name do the `ActionContext`?  It would allow us to query BigQuery at least

leplatrem:
Is this relevant if we do #21 ?

alexcottner:
I don't think this is relevant now that we have 1 webhook for any project, and the retry queue.
We have some pretty good logging in the retry process and sentry alerting from the web service now.
Going to mark this as closed.

PR #918, opened 2024-03-20
Add Dead Letter Queue class, integrate into heartbeat

grahamalama:
Built on the groundwork of #903 (cheers @leplatrem, when I squash this I'll make sure you're a coauthor), this PR adds the classes that allow us to manage our Dead Letter Queue. 

We included a basic heartbeat check in this PR to assert that we are able to read / write from the queue. I removed the integration with the app (in the `/bugzilla_webhook` route) for now. The idea is that we'll deploy this change first to validate that we can connect to the directory we're mounting on the infra side. Then, we'll follow on with how we integrate it into the `/bugzilla_webhook` route.


alexcottner (approved):
looking good

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
+"""This `queue` module stores Bugzilla webhook messages that we failed to sync
+to Jira.
+
+As Bugzilla sends us webhook messages, we want to eagerly accept them and
+return a `200` response so that we don't prevent it from sending new messages.
+But if we fail to sync a bug, we want to keep the message so we can retry it
+later. We also want to store any messages that might be successfuly synced, but
+were preceded by a message that wasn't synced.
+
+Classes:
```
leplatrem, line 10: Is that a specific docstring format?

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+    @abstractmethod
+    async def get(self, bug_id: int) -> list[QueueItem]:
+        """Retrieve all of the queue items for a specific bug, sorted in
+        ascending order by the timestamp of the payload event.
+        """
+        pass
+
+    @abstractmethod
+    async def get_all(self) -> dict[int, list[QueueItem]]:
+        """Retrive all items in the queue, grouped by bug
```
leplatrem, line 116:
```suggestion
        """Retrieve all items in the queue, grouped by bug
```

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+
+class MemoryBackend(QueueBackend):
+    def __init__(self):
+        self.existing: dict[int, list[QueueItem]] = defaultdict(list)
+
+    def ping(self):
+        return True
+
+    async def clear(self):
+        logger.debug("Clearing queue")
```
leplatrem, line 139: I think we want to avoid `debug()` level. In production they are very likely to be ignored. And our volume of application logging, compared to k8s etc., is ridiculously small, so we can be generous and verbose here I think

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+        if bug_id not in self.existing:
+            return []
+        items = self.existing[bug_id]
+        if not items:
+            logger.warn("No items for bug %s, but present in queue", bug_id)
+        return items
+
+    async def get_all(self) -> dict[int, list[QueueItem]]:
+        empty_entries = [not items for items in self.existing.values]
+        if empty_entries:
```
leplatrem, line 161: You could use `if empty_entries := [...]`: 😊

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+        # this way so that we don't create a key for a bug with no items
+        if bug_id not in self.existing:
+            return []
+        items = self.existing[bug_id]
+        if not items:
+            logger.warn("No items for bug %s, but present in queue", bug_id)
+        return items
+
+    async def get_all(self) -> dict[int, list[QueueItem]]:
+        empty_entries = [not items for items in self.existing.values]
```
leplatrem, line 160: isn't `values()` 🤔 ?

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+            )
+        return self.existing
+
+    async def remove(self, bug_id: int, identifier: str):
+        filtered_items = [
+            i for i in self.existing[bug_id] if i.identifier != identifier
+        ]
+        if not len(filtered_items):
+            del self.existing[bug_id]
+            logger.debug(
```
leplatrem, line 173: ditto about `debug`

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+            return True
+        except Exception:
+            logger.exception("Could not write to file backed queue")
+            return False
+
+    async def clear(self):
+        shutil.rmtree(self.location)
+
+    async def put(self, item: QueueItem):
+        folder = self.location / f"{item.payload.bug.id}"
```
leplatrem, line 204: I wondering... is it more idiomatic to write `str({item.payload.bug.id)` instead?

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+    def __init__(self, location):
+        self.location = Path(location)
+        self.location.mkdir(parents=True, exist_ok=True)
+
+    def ping(self):
+        try:
+            with tempfile.TemporaryDirectory(dir=self.location) as temp_dir:
+                with tempfile.TemporaryFile(dir=temp_dir) as f:
+                    f.write(b"")
+            return True
```
leplatrem, line 195: I was thinking, maybe we could also check `self.get_all()`. Because if a malformed JSON file cannot be parsed, then we want to be alerted

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        Returns the whole list of items in the queue, grouped by bug, and
+        sorted from oldest to newest.
+        """
+        existing = await self.backend.get_all()
+        # Sort by event datetime ascending.
+        by_oldest_bug_events = sorted(
+            existing.values(), key=lambda items: items[0].payload.event.time or 0
+        )
+        return list(itertools.chain(*by_oldest_bug_events))
```
leplatrem, line 309: Like I was [noting in #903](https://github.com/mozilla/jira-bugzilla-integration/pull/903#discussion_r1521286965)

After thinking about this, we shouldn't flatten this list. But if the scheduler fails to process one request, it will want to jump to the next bug. 

```python

for items in all_items:
    for item in items:
        try:
            execute_action(item.payload, actions)
        except IgnoreInvalidRequestError:
            await queue.done(item)
        except Exception as exc:
            # Jump to next bug
            break
        else:
            await queue.done(item)
```

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+class PythonException(BaseModel, frozen=True):
+    type: str
+    description: str
+    details: str
+
+    @classmethod
+    def from_exc(cls, exc: Exception):
+        return PythonException(
+            type=exc.__class__.__name__,
+            description=str(exc),
```
leplatrem, line 57: As said in #903, since `HTTPError` are very opaque, for easier debugging I think we could dump the response status (and body?) for `HTTPError`, with a `isinstance(exc, HTTPError) and hasattr(exc, "response")`

leplatrem (commented):
Thank you Graham for bringing this to another level 🙏 💯

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+        return f"{self.timestamp.isoformat()}-{self.payload.bug.id}-{self.payload.event.action}-{"error" if self.error else "postponed"}"
+
+
+@lru_cache(maxsize=1)
+def get_dl_queue():
+    settings = get_settings()
+    return DeadLetterQueue(settings.dl_queue_dsn)
+
+
+class QueueBackend(ABC):
```
alexcottner, line 80: Looking at this with less sinus pressurized eyes today: Is this too much abstraction? If we're always going to be running a FileBackend in deployed services, should we even allow a MemoryBackend to be used locally?  

This feels like we might be making it easier for our future selves to make mistakes and not catch them until they're deployed.

jbi/queue.py
```diff
@@ -0,0 +1,315 @@
...
+
+    async def clear(self):
+        shutil.rmtree(self.location)
+
+    async def put(self, item: QueueItem):
+        folder = self.location / f"{item.payload.bug.id}"
+        folder.mkdir(exist_ok=True)
+        path = folder / (item.identifier + ".json")
+        path.write_text(item.model_dump_json())
+        logger.debug("%d items in dead letter queue", self.size)
```
alexcottner, line 208: self.size will be executed regardless of log level, correct? I don't think python does any magic here to not do that.

In that case, this may be problematic with a busy FileBackend. But we _shouldn't_ have that.

jbi/queue.py
grahamalama: 🤦🏻‍♂️ [9bd9c96](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/9bd9c96561d8d3149310037c84a26227e5c80044)

jbi/queue.py
grahamalama: [9bd9c96](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/9bd9c96561d8d3149310037c84a26227e5c80044)

jbi/queue.py
grahamalama: Yeah I was thinking that when I first saw it. I'm not sure?

jbi/queue.py
grahamalama: > self.size will be executed regardless of log level, correct?

Nope, size will be lazily computed for this log line only when the log level is set to debug

jbi/queue.py
grahamalama: Nope. I was trying to find if there was one. My goal was to document a "TL:DR" of the module + a high level look at the classes. If you have better formatting ideas, I'm all ears!

jbi/queue.py
grahamalama: Yeah, I think this is the right call / something for you to take into account with the retry script. @alexcottner , what do you think?

jbi/queue.py
grahamalama: > Is this too much abstraction?

Hmm. Yes, I think you're right. Now that I'm thinking about it, having multiple backend classes might make more sense for something like a library or if this queue was used by multiple apps with different backend types. However:

> If we're always going to be running a FileBackend in deployed services

I don't know why, but some part of me thinks we'll be moving away from our filesystem solution eventually. If / when we do, it'll be nice to rely on the API for DeadLetterQueue staying stable. 

To that end, I do think it's nice to have this interface regardless of our backend(s). It helps `DeadLetterQueue` stay stable by keeping us honest with how it expects to interact with a backend.

So I think we should have 3 classes:
- DeadLetterQueu
- the QueueBackend interface
- the FileBackend class

What do you think @leplatrem @alexcottner?

jbi/queue.py
grahamalama: Things brings up a good point that currently one "bad apple" can spoil the whole queue. We should probably account for this, especially since one of the reasons we've rejected BZ messages in the past was because they didn't conform to our model.

Maybe we could return a dict of `{int, Generator[QueueItem]}` (or whatever the right syntax for generators is)? So when the retry script hits one of these malformed payloads, we can catch the exception and move on to the next bug. This prevents us from knowing this size of the queue though, I think

grahamalama:
One more big though I'm having about the architecture of this PR:

We've been calling this _a_ (singular) queue. But almost feels like we're building a queue _manager_. Like to me, what we return from `get_all()` is not a queue, but a collection of queues. And each bug has its own queue of failing messages. Thoughts?

jbi/queue.py
alexcottner: If that makes us feel more confident about changing things in the future, I'm good with that. That's at least dropping the in-memory queue, which is something we don't have plans for using in a deployed environment. 

I try to focus on the problem in front of me and solve exactly that, but without painting myself into a corner. So as long as it's easy to pivot in the future and we aren't doing a bunch of extra work now, I'll be happy.

alexcottner:
> We've been calling this _a_ (singular) queue. But almost feels like we're building a queue _manager_. Like to me, what we return from `get_all()` is not a queue, but a collection of queues. And each bug has its own queue of failing messages. Thoughts?

In my head, this is a very simple partitioned queue. We have a single queue where things are going, but blocking only occurs within the partition. These terms would be more well defined if we were using an actual queue service with partitions and compute nodes.

jbi/queue.py
leplatrem: This is good ! I was just curious

jbi/queue.py
leplatrem: This does not matter too much :)

jbi/queue.py
leplatrem: I originally added the memorybackend for the tests, to be sure we start from a clean state etc.
But I'm definitely +1 on removing it for the reasons you mention. Keep the interface to maintain the shape of the composition pattern also seems a wise choice 💯

jbi/queue.py
leplatrem: Good idea for the generator. I hadn't thought that it could delegate the exception handling to the consumer.

The size of queue could be return by the backend just counting files, without parsing them.

I believe we really should know the size of the queue, in order to add warning logs when the size is over a configured threshold (eg. 100?)

The heartbeat endpoint would:
- check that the folder is writable
- iterate the whole queue to make sure all events are parseable
- warn if the size of the queue is over a threshold (100?)
- warn if the age of the oldest event is over a threshold (2 days?)

Then we can also imagine high-levels logs for easier debugging (could be part of retry job?)
- log the time window of events (eg. `Queue contains 14 events starting from 20240301`) 
- log the list of projects concerned by the queued events (eg. `Queue contains 14 events for projects FIDDEDE, FXP`), related #567

grahamalama: > The size of queue could be return by the backend just counting files, without parsing them.

Yep, good call. I've accounted for that in my latest commits 👍🏻

jbi/queue.py
```diff
@@ -0,0 +1,258 @@
...
+    async def track_failed(
+        self, payload: bugzilla.WebhookRequest, exc: Exception
+    ) -> None:
+        """
+        Store the specified payload and exception information into the queue.
+        """
+        item = QueueItem(
+            payload=payload,
+            error=PythonException.from_exc(exc),
+        )
```
leplatrem, line 236: Should we be smart here and increment the `retries` field if the request is already in the queue?

Or should we remove the `retries` field for now?

alexcottner: Thinking about it more, I'm in favor of removing retries property. We can calculate how many times we've retried a given entry or look at the logs if we need to. Updating it each time will cause more writes to the file share, and increase the risk of a file getting corrupted during a write (although that should be a very low risk).

jbi/queue.py
grahamalama: To be covered in a followup, based on your approval message

tests/unit/jira/test_queue.py
```diff
@@ -0,0 +1,196 @@
...
+    await backend.put(item_1)
+    await backend.put(item_3)
+    await backend.put(item_4)
+
+    items = [item async for item in backend.get(item_1.payload.bug.id)]
+    assert list(items) == [item_1, item_2, item_3, item_4]
+
+
+@pytest.mark.asyncio
+async def test_get_all(backend: QueueBackend, queue_item_factory):
```
alexcottner, line 87: Should we add a `test_get_all_failed` test where one file has invalid JSON in it?

alexcottner (approved):
This is looking really good. Left comment about an additional possible test scenario.

tests/unit/jira/test_queue.py
grahamalama: Good call. I added a few tests here [d6610f3](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/d6610f316a30c40d3c424e2b270a47d34db2cfd5).

Should we catch at reraise these as our own QueueError or something?

leplatrem: I'd say yes, just for the Slack notification to be explicit (ie. validation error can apply to incoming request or queue content now)

grahamalama: 👍🏻 [43a7a15](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/43a7a158b6c80207fef65d6d0e66c6caa1234cdf)

jbi/queue.py
```diff
@@ -0,0 +1,302 @@
...
+                all_items[int(filesystem_object.name)] = self.get(filesystem_object)
+        return all_items
+
+    async def size(self, bug_id=None) -> int:
+        location = self.location / str(bug_id) if bug_id else self.location
+        return sum(1 for _ in location.rglob("*.json"))
+
+
+class InvalidQueueDSNError(Exception):
+    pass
```
leplatrem, line 235: Move this to top along with the other?

leplatrem (approved):

jbi/queue.py
```diff
@@ -167,13 +196,31 @@ async def remove(self, bug_id: int, identifier: str):
...
         if not folder.is_dir():
             return
             yield
         for path in sorted(folder.iterdir()):
-            yield QueueItem.parse_file(path)
+            try:
+                yield QueueItem.parse_file(path)
+            except (JSONDecodeError, ValidationError) as e:
+                raise QueueItemRetrievalError(
+                    "Unable to load item at path %s from queue", str(path)
```
alexcottner, line 222: 🤘

PR #924, opened 2024-03-26
Adding retry process for dead letter queue

alexcottner:
Built on top of #918 .

This adds the retry process and associated unit tests to reprocess failed events from the Dead Letter Queue. 
Added unit tests for scenarios outlined in [ADR](https://github.com/alexcottner/jira-bugzilla-integration/blob/adr-003/docs/adrs/003-bugzilla-response-codes.md).


tests/unit/test_retry.py
```diff
@@ -0,0 +1,164 @@
...
+
+    await retry_failed()
+    retrieve.assert_called_once()
+    logger.info.assert_not_called()
+    logger.warn.assert_not_called()
+    logger.error.assert_not_called()
+
+
+def add_iter(obj):
+    mock = MagicMock()
```
alexcottner, line 47: This really feels like I'm doing something the hard way.

jbi/retry.py
```diff
@@ -0,0 +1,61 @@
...
+from os import getenv
+from time import sleep
+from typing import AsyncIterator
+
+import jbi.runner as runner
+from jbi.configuration import ACTIONS
+from jbi.queue import QueueItem, get_dl_queue
+
+CONSTANT_RETRY = getenv("CONSTANT_RETRY", "false") == "true"
+RETRY_TIMEOUT_DAYS = getenv("RETRY_TIMEOUT_DAYS", 7)
```
grahamalama, line 13: If we're specifying these values in `jbi/environment.py`, we should probably access them with
```
settings = get_settings()
...
settings.constant_retry
...
settings.retry_timeout_days
```

However, if these settings only pertain to this script (and won't matter to the FastAPI app), maybe we can move them out of `environment.py` and access them like you have here. That, or build a new `BaseSettings` object so we get validation.

jbi/retry.py
```diff
@@ -0,0 +1,61 @@
...
+
+            try:
+                runner.execute_action(item.payload, ACTIONS)
+                await queue.done(item)
+            except Exception as ex:
+                # write well formed log that could be alerted on
+                logger.error(
+                    "failed to reprocess event %s. error: %s", item.identifier, ex
+                )
+                prev_failed = True
```
grahamalama, line 49: Instead of using a `prev_failed` sentinel, could we `break` here if we encounter an exception? This will take us to the next bug in the queue. If we want to keep the info log, we could do that before we break and say something like `logger.info(skipping remaining events for bug %s, bugid)`

jbi/retry.py
```diff
@@ -0,0 +1,61 @@
...
+            # skip and delete item if we have exceeded max_timeout
+            if item.timestamp < min_event_timestamp:
+                logger.warn("removing expired event %s", item.identifier)
+                await queue.done(item)
+                continue
+
+            try:
+                runner.execute_action(item.payload, ACTIONS)
+                await queue.done(item)
+            except Exception as ex:
```
grahamalama, line 44: Should we split `runner.execute_action(item.payload, ACTIONS)` and `await queue.done(item)` into separate `try / except` blocks? There's a small but non-0 chance that we successfully execute an action, but then when we try to mark the item as done, something goes wrong.

We could even retry marking the item as done if something goes wrong.

tests/unit/test_retry.py
```diff
@@ -0,0 +1,164 @@
...
+import jbi.runner
+from jbi.bugzilla.models import Bug, WebhookEvent, WebhookRequest
+from jbi.queue import QueueItem, get_dl_queue
+from jbi.retry import retry_failed
+
+RETRY_TIMEOUT_DAYS = getenv("RETRY_TIMEOUT_DAYS", 7)
+
+
+@pytest.fixture()
+def logger():
```
grahamalama, line 18: Pytest's [caplog](https://docs.pytest.org/en/6.2.x/logging.html#caplog-fixture) fixture could be useful here

tests/unit/test_retry.py
```diff
@@ -0,0 +1,164 @@
...
+    logger.error = MagicMock()
+    logging.getLogger = lambda x=None: logger
+    return logger
+
+
+@pytest.fixture()
+def execute_action():
+    _execute_action = MagicMock()
+    jbi.runner.execute_action = _execute_action
+    return _execute_action
```
grahamalama, line 31: Maybe you could add a parameter to retry_failed so it's 
```python
def retry_failed(item_executor=runner.execute_action)
```

So that it tests you can do:

```python
retry_failed(item_executor=MagicMock())
```

grahamalama (commented):
Made a few comments as a first pass, I'll think more about how to best manage the queue in another review.

jbi/retry.py
alexcottner: That is how I had the code originally as it's definitely more efficient. But then we don't log anything for the events we're skipping. I started to write something that would log "skipping X future events for bug Y", but that seemed like a truncation of data. So then I went with this.

This also helped with the summary metrics we're logging at the end of the run.

With all of that said, is there a reasonable way to break early and still produce meaningful logs and metrics?

grahamalama: > "skipping X future events for bug Y"

I think this could be enough, but also maybe this means we should add a 

```python
def list(bug_id: Optional[int] = None)
    ...
```

method so that we can return the identifiers in the queue without trying to load them. That way, we could do something like:

```python
try:
    runner.execute_action(item.payload, ACTIONS)
    metrics["events_processed"] += 1
    await queue.done(item)
except Exception as ex:
    logger.exception(
        "failed to reprocess event %s. error: %s", item.identifier, ex
    )
    metrics["events_failed"] += 1
    skipped_items = await queue.list(bugid)
    logger.info(
        "skipping events %s - previous events have failed for this bug",
        ",".join(skipped_items),
    )
    metrics["events_skipped"] += len(skipped_items)
```

grahamalama: [63057fc](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/63057fc3b3be364f6a73fed1428d02997551cfac)

jbi/environment.py
```diff
@@ -48,6 +48,9 @@ class Settings(BaseSettings):
     sentry_dsn: Optional[AnyUrl] = None
     sentry_traces_sample_rate: float = 1.0
 
+    # Retry queue
+    constant_retry: str = "false"
```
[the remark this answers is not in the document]
alexcottner: Removing from settings as suggested elsewhere as this only impacts the retry script.

jbi/retry.py
alexcottner: Mapping this out (this comment might get messy).

Current code workflow looks like this:
1. Process event successfully via `execute_action`
2. Fail to mark as done (can't delete file, likely IO error or maybe a k8s config issue)
3. Skip processing future events for this bug
4. On next retry, we reprocess the same event even though we processed it successfully the first time
5. Mark it as done successfully this time
6. Process other events in queue for this bug as normal

On the surface, that doesn't seem like a problem. We might process the same event twice (several hours apart), but ultimately we're updating the same data to the same values. What problems may occur here that I'm not thinking of?


If we alter the code to note that we failed marking the item as done it would look like:
1. Process event successfully via `execute_action`
2. Fail to mark as done
3. Update the existing file (rename or rewrite) to note that we failed to delete it. Or maybe write an additional file?
4. Continue processing other events for this bug since we didn't actually fail to update
5. On next retry, check for this note so we know to skip `execute_action` and just try to mark it as done
6. Continue as normal

I think this adds more complexity (more branching, more test cases) and is likely to run into the same IO issue that prevented us from deleting the file. I don't think we gain anything, unless we might stop ourselves from overwriting something in Jira that a user would have changed manually.

Does that stream of thoughts seem right? Am I missing anything?

tests/unit/test_retry.py
alexcottner: Oh yeah that's much easier to work with

tests/unit/test_retry.py
alexcottner: I'll make that change. Is this the "python way" of doing things? I'm used to using a lot of dependency injection for tests in other languages, but was treating python more like the looser languages (like node) and mocking them directly.

jbi/retry.py
```diff
@@ -0,0 +1,61 @@
...
+
+async def retry_failed():
+    queue = get_dl_queue()
+    logger = logging.getLogger(__name__)
+    min_event_timestamp = datetime.now() - timedelta(days=int(RETRY_TIMEOUT_DAYS))
+
+    # load all bugs from DLQ
+    bugs: dict[int, AsyncIterator[QueueItem]] = await queue.retrieve()
+
+    for bugid, items in bugs.items():
```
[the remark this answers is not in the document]
alexcottner: updated for consistency

jbi/retry.py
alexcottner: I ran with moving them out of `environment.py` since they're only used in one place. This feels a little weird because we have 2 entry points into the same module. But we're sharing so much code I think it makes sense. Let me know if you think going another direction is better.

.env.example
```diff
@@ -6,6 +6,8 @@ APP_RELOAD=True
 APP_DEBUG=True
 JBI_API_KEY="fake_api_key"
 DL_QUEUE_DSN=file:///tmp
+CONSTANT_RETRY="true"
+RETRY_TIMEOUT_DAYS=7
```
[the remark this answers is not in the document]
alexcottner: Added

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+        "events_processed": 0,
+        "events_skipped": 0,
+        "events_failed": 0,
+        "bugs_failed": 0,
+    }
+
+    for bug_id, items in bugs.items():
+        try:
+            async for item in items:
+                # skip and delete item if we have exceeded max_timeout
```
leplatrem, line 36: nit
```suggestion
                # skip and delete item if we have exceeded RETRY_TIMEOUT_DAYS
```

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+                    metrics["events_failed"] += 1
+
+                    # check for other events that will be skipped
+                    skipped_events = await queue.list(bug_id)
+                    if (
+                        len(skipped_events) > 1
+                    ):  # if this isn't the only event for the bug
+                        logger.info(
+                            "skipping events %s - previous event %s failed for this bug",
+                            ",".join(skipped_events),
```
leplatrem, line 61: How is it going to show up? Especially if many items?
Maybe logging `len(skipped_events)` would be enough here? Or thus get the size with `await queue.size(bug_id)`

jbi/retry.py
leplatrem: > What problems may occur here that I'm not thinking of?

Some operations are creation, like comments for example. 
We may end up with a comment being posted twice. I don't think it's critical and I value simple code over this (especially because once this is setup, failing to write on a disk should not happen easily)

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+    mock = MagicMock()
+    mock.__aiter__.return_value = None
+    mock.__aiter__.side_effect = Exception("Throwing an exception")
+    return mock
+
+
+@pytest.mark.asyncio
+async def test_retry_empty_list(caplog):
+    retrieve = AsyncMock(return_value={})
+    get_dl_queue().retrieve = retrieve
```
leplatrem, line 30: In other tests we mock the instantiation of the singleton in a fixture:

https://github.com/mozilla/jira-bugzilla-integration/blob/dca677f6ecb94607c3468da85a2477283833e5a3/tests/conftest.py#L110-L118

Maybe we want to follow the same pattern?

I think there is also an opportunity to leverage the composition pattern on the queue backend.
We could chose to not mock the queue methods but inject a mocked backend instead.

(completely untested)
```py
@pytest.fixture()
def mocked_queue(request):
    class MockedBackend(QueueBackend):
         get_all = AsyncMock()
         size = MagicMock()
         ...
    
    with mock.patch("jbi.queue.get_dl_queue") as mocked:
        mocked.return_value.backend = MockedBackend()
        yield
    jbi.queue.get_dl_queue.cache_clear()

```

AFAIU, the backend methods could be have their return value set like this:

```py
async def test_retry_empty_list(caplog, mocked_queue):
     mocked_queue.backend.get_all.return_value=[]
```

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+                    )
+                ]
+            )
+        }
+    )
+    queue.done = AsyncMock()
+
+    metrics = await retry_failed(item_executor=exec_mock)
+    assert len(caplog.messages) == 0
+    queue.retrieve.assert_called_once()  # no logs should have been generated
```
leplatrem, line 63: nit: I think the trailing comment goes to the line above

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+                    queue_item_factory(
+                        payload__bug__id=1, payload__event__time=datetime.now()
+                    ),
+                ]
+            )
+        }
+    )
+
+    exec_mock = MagicMock()
+    exec_mock.side_effect = Exception("Throwing an exception")
```
leplatrem, line 94: nit: you can do this in one line passing the `side_effect` to the constructor

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+    }
+    queue.retrieve = AsyncMock(return_value=mock_data)
+
+    queue.done = AsyncMock()
+    metrics = await retry_failed(item_executor=exec_mock)
+    queue.retrieve.assert_called_once()
+    assert (
+        len(queue.done.call_args_list) == 2
+    )  # both items should have been marked as done
```
leplatrem, line 139: nit: you can put this message as part of the `assert`

```suggestion
    assert (
        len(queue.done.call_args_list) == 2
    ), "both items should have been marked as done"
```

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+
+@pytest.mark.asyncio
+async def test_retry_bug_failed(caplog, queue_item_factory):
+    exec_mock = MagicMock()
+    queue = get_dl_queue()
+    mock_data: dict[int, AsyncIterator[QueueItem]] = {
+        1: add_iter(
+            [
+                queue_item_factory(
+                    payload__bug__id=1, payload__event__time=datetime.now()
```
leplatrem, line 161: nit: If we always use `datetime.now()` then maybe this should be the default in the factory

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+        "bugs_failed": 0,
+    }
+
+    for bug_id, items in bugs.items():
+        try:
+            async for item in items:
+                # skip and delete item if we have exceeded max_timeout
+                if item.timestamp < min_event_timestamp:
+                    logger.warning("removing expired event %s", item.identifier)
+                    await queue.done(item)
```
leplatrem, line 39: Do we need another method for this? `discard()` or something with better semantics than `done()`? Especially if in the queue we log stuff like `item X is done`

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+async def main():
+    logger = logging.getLogger(__name__)
+    while True:
+        metrics = await retry_failed()
+        logger.info("event queue processing complete", extra=metrics)
+        if not CONSTANT_RETRY:
+            return
+        sleep(5)
```
leplatrem, line 80: nit: put this `5` into a configurable constant

tests/unit/test_retry.py
```diff
@@ -0,0 +1,184 @@
...
+from os import getenv
+from typing import AsyncIterator
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from jbi.queue import QueueItem, get_dl_queue
+from jbi.retry import retry_failed
+
+RETRY_TIMEOUT_DAYS = getenv("RETRY_TIMEOUT_DAYS", 7)
```
leplatrem, line 11: Don't you want to import this from `jbi.retry` ?

tests/unit/test_retry.py
grahamalama: In addition to these good suggestions, I tried playing around with this and came up with:


```python
@pytest.fixture
def queue():
    return MagicMock(spec=DeadLetterQueue)


@pytest.fixture
def executor():
    return MagicMock(spec=execute_action)


async def aiter_sync(iterable):
    for i in iterable:
        yield i


@pytest.mark.asyncio
async def test_retry_success(caplog, queue, queue_item_factory, executor):
    queue.retrieve.return_value = {1: aiter_sync([queue_item_factory()])}
    metrics = await retry_failed(queue, executor)
    queue.retrieve.assert_called_once()  # no logs should have been generated
    queue.done.assert_called_once()  # item should be marked as complete
    executor.assert_called_once()  # item should have been processed
    assert metrics == {
        "bug_count": 1,
        "events_processed": 1,
        "events_skipped": 0,
        "events_failed": 0,
        "bugs_failed": 0,
    }
```

It's more dependency injection, but it's nice because we're explicitly building an object with behavior we expect for the test, then passing it directly to the function under test

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+                    metrics["events_processed"] += 1
+                except Exception as ex:
+                    # write well formed log that could be alerted on
+                    logger.error(
+                        "failed to reprocess event %s. error: %s", item.identifier, ex
+                    )
+                    metrics["events_failed"] += 1
+
+                    # check for other events that will be skipped
+                    skipped_events = await queue.list(bug_id)
```
grahamalama, line 55: A mocking gotcha. Somehow we didn't catch that `queue.list` isn't a method (an oversight on my part), rather it's `queue.backend.list`. I think creating a MagicMock with [`spec`](https://docs.python.org/3/library/unittest.mock.html#auto-speccing). will help us avoid things like this.

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+from jbi.configuration import ACTIONS
+from jbi.queue import QueueItem, get_dl_queue
+
+CONSTANT_RETRY = getenv("DL_QUEUE_CONSTANT_RETRY", "false") == "true"
+RETRY_TIMEOUT_DAYS = getenv("DL_QUEUE_RETRY_TIMEOUT_DAYS", 7)
+
+
+async def retry_failed(item_executor=runner.execute_action):
+    queue = get_dl_queue()
+    logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
leplatrem: 👍 and remove the other at the bottom

jbi/retry.py
grahamalama: [1ee3ab7](https://github.com/mozilla/jira-bugzilla-integration/pull/918/commits/1ee3ab7f921311e094a0b5ce5211730c40097fde)

tests/unit/test_retry.py
alexcottner: Yup, got misaligned when switching to caplog

jbi/retry.py
alexcottner: The queue should be just an IO operator and not have any real logic in it. It feels a little weird to have two different functions that do the same functional operation but with potentially different debug logs. Am I thinking about it wrong?

Looking at the logs the queue emits, we don't currently have a conflict problem. The debug level logs say things like "Removed {event} from queue for bug {bug_id}." and "Removed directory for bug {bug_id}". And our log levels will be above debug in prod and we won't alert on those.

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+                    logger.error(
+                        "failed to reprocess event %s. error: %s", item.identifier, ex
+                    )
+                    metrics["events_failed"] += 1
+
+                    # check for other events that will be skipped
+                    skipped_events = await queue.list(bug_id)
+                    if (
+                        len(skipped_events) > 1
+                    ):  # if this isn't the only event for the bug
```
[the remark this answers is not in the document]
alexcottner: The current event we're processing will exist in the queue, so `len(queue.list(bug_id))` will always be at least 1. We are only skipping other events if there are more than 1 event pending for the current bug. Does that makes sense and can I make this comment clearer?

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+from jbi.queue import QueueItem, get_dl_queue
+
+CONSTANT_RETRY = getenv("DL_QUEUE_CONSTANT_RETRY", "false") == "true"
+RETRY_TIMEOUT_DAYS = getenv("DL_QUEUE_RETRY_TIMEOUT_DAYS", 7)
+
+
+async def retry_failed(item_executor=runner.execute_action):
+    queue = get_dl_queue()
+    logger = logging.getLogger(__name__)
+    min_event_timestamp = datetime.now() - timedelta(days=int(RETRY_TIMEOUT_DAYS))
```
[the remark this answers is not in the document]
alexcottner: Yeah that's a big 10-4, don't want to deal with timezone mismatches.

jbi/retry.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # load all bugs from DLQ
+    bugs: dict[int, AsyncIterator[QueueItem]] = await queue.retrieve()
```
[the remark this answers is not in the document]
alexcottner: I think I had an intellisense issue so I added that to be explicit. But I'm not anymore so removing.

jbi/retry.py
alexcottner: That's a good thing to note. If that becomes an issue, I think the easiest solution is to do upserts for all create/update commands. That might be a little cumbersome through the Jira API but hopefully we can create a simple wrapper for ourselves if it is.

This also fixes the issue where something was created, but didn't make it into Jira (for whatever reason). And then we receive an update for it later.

jbi/retry.py
alexcottner: I figured this would be < 5 events normally, but some bugs can get pretty busy so that's fair. At first I didn't want to truncate this data, but maybe saying "5 events skipped" is good enough for our logging/alerting.

jbi/retry.py
alexcottner: Added as an environment var, defaulting to 5.

jbi/retry.py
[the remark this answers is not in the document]
alexcottner: Switched this to just pull `size()`. But same deal, it will count the current event so we need to subtract 1.

tests/unit/test_retry.py
```diff
@@ -0,0 +1,158 @@
...
+
+@pytest.fixture
+def mock_executor():
+    return MagicMock(spec=execute_action)
+
+
+@pytest.mark.asyncio
+async def test_retry_empty_list(caplog):
+    retrieve = AsyncMock(return_value={})
+    get_dl_queue().retrieve = retrieve
```
leplatrem, line 36: Why don't we use `mock_queue` here like in others?

leplatrem (approved):

PR #938, opened 2024-04-09
Group pip updates pull-requests and add GH Actions updates

leplatrem:


.github/dependabot.yml
```diff
@@ -4,7 +4,14 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
```
alexcottner, line 6: Should we change this to weekly now?

.github/dependabot.yml
```diff
@@ -4,7 +4,14 @@ updates:
...
+    groups:
+      minor-patch:
+        update-types: ["minor", "patch"]
```
alexcottner, line 9: Should we include `"major"` too?

alexcottner (approved):

PR #939, opened 2024-04-10
Adding queue to bugzilla webhook route

alexcottner:
This PR aims to finish the retry queue work by adding logic to the `bugzilla_webhook` route

- [x]  Adds queue logic to the bugzilla webhook route
- [x]  Adds additional unit tests to cover branches
- [x]  Adds well formed logs for easy alerting

Much of this was copied from @leplatrem 's [draft PR here](https://github.com/mozilla/jira-bugzilla-integration/pull/903).


tests/unit/test_runner.py
```diff
@@ -9,7 +9,13 @@
...
 from jbi.errors import ActionNotFoundError, IgnoreInvalidRequestError
 from jbi.models import ActionContext
-from jbi.runner import Executor, execute_action, lookup_action
+from jbi.queue import DeadLetterQueue
+from jbi.runner import Actions, Executor, execute_action, execute_or_queue, lookup_action
+
+
+@pytest.fixture
+def mock_queue():
+    return mock.MagicMock(spec=DeadLetterQueue)
```
leplatrem, line 18: This could probably go to conftest if it's used across different modules

tests/unit/test_runner.py
```diff
@@ -249,6 +255,57 @@ def test_runner_ignores_request_if_jira_is_linked_but_without_whiteboard(
...
+@pytest.mark.asyncio
+async def test_execute_or_queue_happy_path(
+    mock_queue, 
+    bugzilla_webhook_request,
+):
+    mock_queue.is_blocked.return_value = False
+    await execute_or_queue(
+        request=bugzilla_webhook_request,
+        queue=mock_queue,
+        actions=mock.MagicMock(spec=Actions)
```
leplatrem, line 267: why not use the `actions` fixture?

tests/unit/test_runner.py
```diff
@@ -249,6 +255,57 @@ def test_runner_ignores_request_if_jira_is_linked_but_without_whiteboard(
...
+async def test_execute_or_queue_blocked(
+    actions, 
+    mock_queue, 
+    bugzilla_webhook_request,
+):
+    mock_queue.is_blocked.return_value = True
+    await execute_or_queue(
+        request=bugzilla_webhook_request,
+        queue=mock_queue,
+        actions=mock.MagicMock(spec=Actions)
```
leplatrem, line 284: you have the `actions` fixture available in function 😉

tests/unit/test_runner.py
```diff
@@ -249,6 +255,57 @@ def test_runner_ignores_request_if_jira_is_linked_but_without_whiteboard(
...
+    # should trigger an exception for this scenario
+    await execute_or_queue(
+        request=bugzilla_webhook_request,
+        queue=mock_queue,
+        actions=actions
+    )
+    mock_queue.is_blocked.assert_called_once()
+    mock_queue.postpone.assert_not_called()
+    mock_queue.track_failed.assert_called_once()
+
```
leplatrem, line 307: Can we add a test that specifies that we return a 5XX when `queue.track_failed()` or `queue.postpone()` raises?

jbi/runner.py
```diff
@@ -157,6 +158,29 @@ def __call__(self, context: ActionContext) -> ActionResult:
         return True, {"responses": responses}
 
 
+async def execute_or_queue(
+    request: bugzilla.WebhookRequest, queue: DeadLetterQueue, actions: Actions
+):
+    if await queue.is_blocked(request):
```
grahamalama, line 164: I think we should think about gracefully handling when the queue is down, somehow. 

For short-term outages, I think it's enough to retry checking if the queue is blocked and/or postponing the bug until we can connect again. But for long-term outages, I'm not sure what we can do 🤔.

For example, say a bug has 2 events in the queue, but also the queue is unavailable for some reason. What should we do with a newly incoming webhook payload for the queued bug? 

JBI can't know about the bugs in the queue if it can't connect to it. So does it hold on to bugs until the queue becomes available and it can process them again? Does it go "even though I might have queued bugs for this incoming webhook, I'm going to try processing them anyway"?

leplatrem: I may not see every aspect of this, but I think this is good enough to return a 5XX if the queue fails.

If we return a 5XX we are in the same situation as we are now when we fail at processing the event: bugzilla will retry. And eventually disable the hook if fails for a long time, but that would be a service disruption, hence consistent.

tests/unit/test_runner.py
alexcottner: Moving it to conftest now that the other PR has been merged.

tests/unit/test_runner.py
alexcottner: Looking at the other tests that use the `actions` fixture, that would require mocking deeper things (like get_bug and get_issue) right? Since we're just testing the `execute_or_queue` logic here this seemed like the path of least resistance.

tests/unit/test_runner.py
alexcottner: Adding in next commit.

jbi/runner.py
alexcottner: Adjusting so that if the queue has an error we return 500. Adding unit test to verify.

leplatrem (approved):
I would say that we should let the queue crash, and let fastapi return a 500 and send a sentry, instead of catching all exceptions and reraising a httperror manually

.github/labeler.yml
grahamalama, line 1: Let's take this out of this PR and handle it in a separate one. We can merge without having this check pass.

tests/unit/test_router.py
```diff
@@ -156,6 +157,38 @@ def test_webhook_is_200_if_action_raises_IgnoreInvalidRequestError(
...
+
+def test_webhook_is_500_if_queue_raises_Exception(
+    webhook_request_factory, mocked_bugzilla, authenticated_client
+):
+    # queue =
+    # is_blocked = queue.is_blocked
+    webhook = webhook_request_factory()
+    get_dl_queue().is_blocked = mock.AsyncMock(
+        side_effect=Exception("Throwing an exception")
+    )
```
grahamalama, line 182: Since we're testing the route here, I think a nice "FastAPI way" to do this would be to [override the dependency](https://fastapi.tiangolo.com/advanced/testing-dependencies/#testing-dependencies-with-overrides) for this test.

jbi/runner.py
```diff
@@ -157,6 +159,36 @@ def __call__(self, context: ActionContext) -> ActionResult:
...
+            return {"status": "invalid", "error": str(exc)}
+        except Exception as exc:
+            await queue.track_failed(request, exc)
+            return {"status": "failed", "error": str(exc)}
+    except Exception as exc:
+        logger.exception("Queue exception", exc)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Queue Error",
+        )
```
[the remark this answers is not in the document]
alexcottner: Yeah that makes sense. I think I was just trying to have specific logs.

tests/unit/test_router.py
```diff
@@ -156,6 +157,38 @@ def test_webhook_is_200_if_action_raises_IgnoreInvalidRequestError(
...
+    assert response
+    assert response.status_code == 200
+
+
+def test_webhook_is_500_if_queue_raises_Exception(
+    webhook_request_factory, mocked_bugzilla, authenticated_client
+):
+    # queue =
+    # is_blocked = queue.is_blocked
```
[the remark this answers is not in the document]
alexcottner: Ah yeah, I had some half-thoughts left in there

tests/conftest.py
```diff
@@ -94,7 +95,9 @@ def test_api_key():
 @pytest.fixture
 def authenticated_client(app, test_api_key):
     """An test client with a valid API key."""
-    return TestClient(app, headers={"X-Api-Key": test_api_key})
+    return TestClient(
+        app, headers={"X-Api-Key": test_api_key}, raise_server_exceptions=False
```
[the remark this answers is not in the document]
alexcottner: The exception causing the 500 will cause the tests to fail without that. 
```
FAILED tests/unit/test_router.py::test_webhook_is_500_if_queue_raises_Exception - Exception: Throwing an exception
```

tests/unit/test_router.py
alexcottner: Ah that seems easy enough. Adjusting

.github/labeler.yml
alexcottner: Yeah, was hoping it was a quick thing. Reverted.

PR #954, opened 2024-04-18
Adding log formatter to retry process

alexcottner:
Adding JSON log formatter to retry process.


grahamalama (approved):
Make sure to pass `sys.stdout` to the stream handler, then this looks good to go

PR #959, opened 2024-04-19
Fix #958: add more logging fields to retry job exceptions

### LINKED ISSUE #958: Add more context fields to queue exception logs
bug id, etc.

leplatrem:
Fix #958

### DIFF
--- a/jbi/retry.py
+++ b/jbi/retry.py
@@ -52,8 +52,16 @@ async def retry_failed(item_executor=runner.execute_action, queue=get_dl_queue()
                     await queue.done(item)
                     metrics["events_processed"] += 1
                 except Exception:
-                    logger.exception("failed to reprocess event %s.", item.identifier)
                     metrics["events_failed"] += 1
+                    logger.exception(
+                        "failed to reprocess event %s.",
+                        item.identifier,
+                        extra={
+                            "metrics": metrics,
+                            "item": item.model_dump(),
+                            "bug": {"id": bug_id},
+                        },
+                    )
 
                     # check for other events that will be skipped
                     pending_events = await queue.size(bug_id)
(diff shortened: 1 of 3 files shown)


jbi/retry.py
```diff
@@ -52,8 +52,16 @@ async def retry_failed(item_executor=runner.execute_action, queue=get_dl_queue()
                     await queue.done(item)
                     metrics["events_processed"] += 1
                 except Exception:
-                    logger.exception("failed to reprocess event %s.", item.identifier)
                     metrics["events_failed"] += 1
+                    logger.exception(
+                        "failed to reprocess event %s.",
+                        item.identifier,
+                        extra={
+                            "metrics": metrics,
```
grahamalama, line 60: I'm not sure that it's helpful to emit the metrics here. A dump of the item and the bug id make sense though.

jbi/retry.py
```diff
@@ -67,8 +75,12 @@ async def retry_failed(item_executor=runner.execute_action, queue=get_dl_queue()
                         metrics["events_skipped"] += pending_events - 1
                         break
         except Exception:
-            logger.exception("failed to parse events for bug %d.", bug_id)
             metrics["bugs_failed"] += 1
+            logger.exception(
+                "failed to parse events for bug %d.",
+                bug_id,
+                extra={"metrics": metrics, "bug": {"id": bug_id}},
```
grahamalama, line 82: Same here, with the questions about metrics being included in individual log lines

grahamalama (commented):
I added some questions about metrics. Maybe instead of including the metrics in each log line, we can add like a "retry run id" (with a better name), and then include that in each log line. Then we can quickly find all of the logs for a particular retry run (with the metrics appearing on the last line as a summary for that run)

alexcottner:
I'd like to see something added to `execute_or_queue` in `runner.py` as well. If `execute_action` throws an error, we should log an error with the event identifier so we can alert.

jbi/retry.py
alexcottner: I agree, I think logging the metrics multiple times could cause confusion. The rest of this looks good though.

grahamalama (approved):

alexcottner (approved):

PR #961, opened 2024-04-19
Fix #960: add endpoint to view queue content

### LINKED ISSUE #960: Add endpoint to introspect queue
Avoid looking for storage in GCP Console UI

leplatrem:
Fix #960


jbi/router.py
```diff
@@ -75,6 +76,20 @@ async def bugzilla_webhook(
...
+    dependencies=[Depends(api_key_auth)],
+)
+async def inspect_dl_queue(queue: Annotated[DeadLetterQueue, Depends(get_dl_queue)]):
+    """API for viewing queue content"""
+    bugs = await queue.retrieve()
+    results = defaultdict(list)
+    for bug_id, items in bugs.items():
+        async for item in items:
+            results[bug_id].append(item.model_dump())
+    return results
```
grahamalama, line 90: Wouldn't this be a good usecase for the unused code we're deleting in https://github.com/mozilla/jira-bugzilla-integration/pull/964/?

jbi/router.py
```diff
@@ -75,6 +76,20 @@ async def bugzilla_webhook(
...
+@router.get(
+    "/dl_queue/",
+    dependencies=[Depends(api_key_auth)],
+)
+async def inspect_dl_queue(queue: Annotated[DeadLetterQueue, Depends(get_dl_queue)]):
+    """API for viewing queue content"""
+    bugs = await queue.retrieve()
+    results = defaultdict(list)
+    for bug_id, items in bugs.items():
+        async for item in items:
```
grahamalama, line 88: If we're going to read each item here, we need to account for the possibility that the webhook payload somehow doesn't match our schema and throws a validation exception. That's why the current code to retrieve all of the items in the queue only retrieves the identifiers

jbi/router.py
```diff
@@ -75,6 +76,20 @@ async def bugzilla_webhook(
     return await execute_or_queue(webhook_request, queue, actions)
 
 
+@router.get(
+    "/dl_queue/",
```
grahamalama, line 80: Just thinking out loud here, but maybe this `/dl_queue/` route could list all of the items (by identifier only), then `/dl_queue/<identifier>/` could be the detail view of one bug? And/or `/dl_queue/<bug_id>/<identifier>/`

tests/conftest.py
```diff
@@ -164,6 +165,12 @@ def sleepless(monkeypatch):
     monkeypatch.setattr(time, "sleep", lambda s: None)
 
 
+@pytest_asyncio.fixture()
+async def clear_dl_queue():
+    await get_dl_queue().backend.clear()
+    yield
```
grahamalama, line 171: Instead of this, I feel like it might be cleaner to do a dependency override for `get_dl_queue` where we use something like we have in the queue tests. Haven't tested this fully, but something like:

```python
@pytest.fixture
def queue(tmp_path):
    return DeadLetterQueue("file://" + str(tmp_path))

@pytest.fixture
def app(queue):
    app.dependency_overrides[get_dl_queue] = lambda: queue
    return jbi.app.app
```

jbi/router.py
leplatrem: I don't understand

Do you mean that we would want to be able to inspect the invalid items in queue via this endpoint?

jbi/router.py
leplatrem: I find the queue and backend confusing with the current naming and vague docstrings

jbi/router.py
```diff
@@ -75,6 +76,20 @@ async def bugzilla_webhook(
...
+    dependencies=[Depends(api_key_auth)],
+)
+async def inspect_dl_queue(queue: Annotated[DeadLetterQueue, Depends(get_dl_queue)]):
```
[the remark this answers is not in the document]
leplatrem: This endpoint is meant to be used for troubleshooting, so I have doubts about truncating the number of items. Instead I would pick certain fields only instead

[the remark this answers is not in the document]
alexcottner: That should work, but it might still take a while to load all those files from the volume. We'll deal with that if we encounter it though.

jbi/router.py
leplatrem: I think my confusion comes from the fact that we still return 422 when incoming payloads are invalid. Or am I wrong?

jbi/router.py
leplatrem: I would vote to start with something simple like this and iterate if we face loading issues. This API endpoint is not meant to be used publicly anyway. We could also turn it to HTML in case we want to prevent consumers to rely on its format

jbi/router.py
leplatrem: I opened #967 and #968

grahamalama (approved):

PR #974, opened 2024-04-24
Update/dockerflow 2024.04.2, separate queue checks

### LINKED ISSUE #972: Queue dockerflow check error
We're seeing an error on the deployed prod JBI from the dockerflow queue check. Need to investigate and resolve.

Queue appears to be working fine and accessible.

alexcottner:
This PR aims to resolve #972.

Steps taken:
1. Updated dockerflow to 2024.04.2.
2. Enabled async checks
3. Refactored queue heartbeat checks to separate concerns
4. Adjusted unit tests
5. Adjusted queue so it only reads folders that are whole numbers (thanks @leplatrem )
    - note: Tried to do this with glob at first, but I don't think there is a way to do this with glob.


jbi/queue.py
```diff
@@ -240,11 +236,29 @@ def ready(self) -> list[dockerflow.checks.CheckMessage]:
...
 
+    async def check_readable(self) -> list[dockerflow.checks.CheckMessage]:
+        results = []
         try:
-            bugs_items = asyncio.run(self.retrieve())
-            for items in bugs_items.values():
-                asyncio.run(async_iter(items))
+            bugs = await self.retrieve()
+            bug_failed = 0
+            parse_success = 0
```
leplatrem, line 246: Aren't they unused? 🤔

jbi/queue.py
```diff
@@ -240,11 +236,29 @@ def ready(self) -> list[dockerflow.checks.CheckMessage]:
...
+
+            for bug_id, items in bugs.items():
+                try:
+                    async for item in items:
+                        parse_success += 1
+                except Exception as exc:
```
leplatrem, line 252: I wouldn't use a catch all here, but a `QueueItemRetrievalError`

jbi/queue.py
```diff
@@ -205,7 +205,7 @@ async def get(self, bug_id: int) -> AsyncIterator[QueueItem]:
     async def get_all(self) -> dict[int, AsyncIterator[QueueItem]]:
         all_items: dict[int, AsyncIterator[QueueItem]] = {}
         for filesystem_object in self.location.iterdir():
-            if filesystem_object.is_dir():
+            if filesystem_object.is_dir() and re.match("\d", filesystem_object.name): # filtering out temp files from checks
```
leplatrem, line 208: It looks like you didn't cherry-pick the commit because I don't see the unit test from #978

jbi/queue.py
```diff
@@ -240,11 +236,29 @@ def ready(self) -> list[dockerflow.checks.CheckMessage]:
...
                 )
             )
+        return results
 
+    async def check_readable(self) -> list[dockerflow.checks.CheckMessage]:
+        results = []
         try:
-            bugs_items = asyncio.run(self.retrieve())
-            for items in bugs_items.values():
-                asyncio.run(async_iter(items))
```
leplatrem, line 247: `async_iter()` util can be removed

jbi/queue.py
```diff
@@ -240,11 +236,29 @@ def ready(self) -> list[dockerflow.checks.CheckMessage]:
...
+                except Exception as exc:
+                    logger.exception(exc)
+                    bug_failed += 1
+                    results.append(
+                        dockerflow.checks.Error(
+                            f"failed to parse events for bug {str(bug_id)}",
+                            hint="check that parked event files are not corrupt",
+                            id="queue.backend.read",
+                        )
+                    )
```
leplatrem, line 261: instead of this whole block, an alternative would be to add a `bug_id` and `identifier` attributes to the `QueueItemRetrievalError` exception, and just do:

```py
try:
    bug_items = (await self.retrieve()).values()
    [[i async for i in items] for items in bug_items]
except QueueItemRetrievalError as exc:
    results.append(... {exc.bug_id}, {exc.identifier}... )
```

jbi/app.py
```diff
@@ -91,7 +91,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         name="jira.all_project_issue_types_exist",
     )
     checks.register(jira_service.check_jira_pandoc_install, name="jira.pandoc_install")
-    checks.register(queue.ready, name="queue.ready")
+    checks.register(queue.check_ready, name="queue.ready")
```
leplatrem, line 94: we could rename `ready` to `writable`

alexcottner: I like that

jbi/queue.py
alexcottner: Will implement something similar to this. If we fail to parse the file we may not have a valid identifier. But we'll definitely have a file path.

jbi/queue.py
alexcottner: Yeah, I meant to go back and put a `pass` in there.

jbi/queue.py
alexcottner: That's reasonable

jbi/queue.py
alexcottner: Fixed! Will credit you in merge.

jbi/queue.py
alexcottner: Implemented something that I think does what we're after.

jbi/queue.py
```diff
@@ -199,13 +200,16 @@ async def get(self, bug_id: int) -> AsyncIterator[QueueItem]:
                 yield QueueItem.parse_file(path)
             except (JSONDecodeError, ValidationError) as e:
                 raise QueueItemRetrievalError(
-                    f"Unable to load item at path {path} from queue"
+                    f"Unable to load item at path {path} from queue",
```
leplatrem, line 203: I think this is superflous and will come out as: `QueueItemRetrievalError: Unable to load item at path /var/lib/jbi/queue/99999/xxxxxxxxxxxxxxxxxxxxx from queue - path: /var/lib/jbi/queue/99999/xxxxxxxxxxxxxxxxxxxxx.`

Passing `path=path` should be enough

leplatrem (approved):

PR #985, opened 2024-04-30
Logging a warning when we try to remove a file but cannot find it

alexcottner:
Adjusted `queue.remove` to log a warning when the file it's trying to remove cannot be found.
This could happen if we change the item model, specifically around the `identifier` property.

### DIFF
--- a/jbi/queue.py
+++ b/jbi/queue.py
@@ -184,8 +184,12 @@ class FileBackend(QueueBackend):
     async def remove(self, bug_id: int, identifier: str):
         bug_dir = self.location / f"{bug_id}"
         item_path = bug_dir / (identifier + ".json")
-        item_path.unlink(missing_ok=True)
-        logger.debug("Removed %s from queue for bug %s", identifier, bug_id)
+        try:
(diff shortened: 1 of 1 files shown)


jbi/queue.py
```diff
@@ -184,8 +184,12 @@ async def put(self, item: QueueItem):
     async def remove(self, bug_id: int, identifier: str):
         bug_dir = self.location / f"{bug_id}"
         item_path = bug_dir / (identifier + ".json")
-        item_path.unlink(missing_ok=True)
-        logger.debug("Removed %s from queue for bug %s", identifier, bug_id)
+        try:
+            logger.debug("Removing %s from queue for bug %s", identifier, bug_id)
+            item_path.unlink()
+        except FileNotFoundError as exc:
+            logger.warning("Item at not found at path %s", str(item_path), exc)
```
leplatrem, line 191:
```suggestion
            logger.warning("Could not delete missing item at path %s", str(item_path), exc)
```

leplatrem (approved):
nit: log warning message does not mention remove/delete

PR #1000, opened 2024-05-15
Account for ignorable payloads in retry script, try to prevent them in runner

grahamalama:
Somehow, some payloads that we'd normally ignore ended up in our Dead Letter Queue. This PR:
- accounts for such payloads in the retry script when trying to process items in the queue
- tries to prevent such items from making in into the queue by moving the guard to check if the bug has a matching a whiteboard tag before we refresh the bug data

### DIFF
--- a/jbi/retry.py
+++ b/jbi/retry.py
@@ -9,6 +9,7 @@ from dockerflow.logging import JsonLogFormatter, request_id_context
 
 import jbi.runner as runner
 from jbi.configuration import ACTIONS
+from jbi.errors import IgnoreInvalidRequestError
 from jbi.queue import get_dl_queue
 
 CONSTANT_RETRY = getenv("DL_QUEUE_CONSTANT_RETRY", "false") == "true"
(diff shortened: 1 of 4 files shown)


alexcottner (approved):

leplatrem (commented):
> tries to prevent such items from making in into the queue by moving the guard to check if the bug has a matching a whiteboard tag before we refresh the bug data

Do you remember why it was done after? 
Any chance that the whiteboard field wouldn't be available in private bugs or something like that?

grahamalama:
@leplatrem I'm not sure why it was done that way. Seems like more of an oversight than anything

> Any chance that the whiteboard field wouldn't be available in private bugs or something like that?

The [first thing we do in `execute_action`](https://github.com/mozilla/jira-bugzilla-integration/blob/439f0b6513b1f71da11d63ffb625d6299236c3f5/jbi/runner.py#L216C1-L218C78) is check if the bug is private and bail out if it is, so we can assume it'll be a public bug with a whiteboard field

Issue #984, opened 2024-04-29
Adding a patch on Bugzilla creates a meaningless comment on Jira


leplatrem:
On Bugzilla:
<img width="654" alt="Screenshot 2024-04-29 at 13 14 48" src="https://github.com/mozilla/jira-bugzilla-integration/assets/546692/872b338f-5af4-442f-be33-7b70d21b51c5">

On Jira:
<img width="418" alt="Screenshot 2024-04-29 at 12 48 27" src="https://github.com/mozilla/jira-bugzilla-integration/assets/546692/cf180eac-eb64-4866-a8de-908bcd415dbd">

Event payload received from Bugzilla seems to be:
```
{
  "insertId": "g2bib0tfpq2yc5tk",
  "jsonPayload": {
    "Fields": {
      "responses_by_step": {},
      "rid": "8acaa3313a960584958f13264bfc3672",
      "current_step": "create_comment",
      "jira": {
        "labels": null,
        "issue": "XXXXX-4799",
        "project": "XXXXX"
      },
      "operation": "comment",
      "msg": "User comment added to Jira issue XXXXX-4799",
      "event": {
        "user": {
          "login": "phab-bot@bmo.tld",
          "real_name": "Phabricator Automation",
          "id": 600971
        },
        "routing_key": "comment.create",
        "action": "create",
        "time": "datetime.datetime(2024, 4, 9, 15, 39, 56)",
        "changes": null,
        "target": "comment"
      },
      "extra": {},
      "action": {},
      "bug": {
        "cf_fx_points": "---",
        "see_also": [
          "https://mozilla-hub.atlassian.net/browse/XXXXX-4799"
        ],
        "comment": {
          "is_private": false,
          "number": 1,
          "creation_time": "datetime.datetime(2024, 4, 9, 15, 39, 54)",
          "id": 1655555555,
          "body": ""
        },
        "type": "task",
        "component": "Sidebar",
        "resolution": "",
        "product": "Firefox",
        "severity": "--",
        "assigned_to": "user@mozilla.com",
        "priority": "--",
        "flags": [],
        "groups": [],
        "creator": "user@mozilla.com",
        "is_private": null,
        "status": "NEW",
        "summary": "To X and Y",
        "id": 123456,
        "whiteboard": "[XXXXX-sidebar]",
        "keywords": []
      }
    },
  },
  "logName": "projects/moz-fx-jbi-prod-0397/logs/stdout",
[...]

PR #1018, opened 2024-06-19
Ref #984: Simple comment when attachment is added to bug

leplatrem:
Ref #984 

**Note**: will require changes in our Webhook. Attachment events must be enabled (see Bugzilla non prod)

### DIFF
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -121,10 +121,11 @@ class JiraService:
 
         if context.event.target == "attachment":
             routing_key = (
-                context.event.routing_key or ".modified"
-            )  # only for type checking.
+                context.event.routing_key or ".modify"  # only to please type checking.
+            )
             _, verb = routing_key.rsplit(".", 1)
-            formatted_comment = f"*{commenter}* {verb}d an attachment"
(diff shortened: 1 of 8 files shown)


jbi/jira/service.py
```diff
@@ -118,14 +118,22 @@ def add_jira_comment(self, context: ActionContext):
...
-        comment = context.bug.comment
-        assert comment  # See jbi.steps.create_comment()
-        assert comment.body  # Also see jbi.steps.create_comment()
+
+        if context.event.target == "attachment":
+            routing_key = (
+                context.event.routing_key or ".modified"
+            )  # only for type checking.
+            _, verb = routing_key.rsplit(".", 1)
+            formatted_comment = f"*{commenter}* {verb}d an attachment"
```
alexcottner, line 127: If I'm reading this correctly, when a user modifies an attachment this will say something like:
Jane Jetson modifie*dd* an attachment

Maybe a little string replace line to clean up the extra "d" is all we need?

alexcottner (approved):
One small comment but the overall logic looks great

jbi/jira/service.py
leplatrem: It will be `modifyed` which is not great either

PR #1029, opened 2024-07-17
Add ability to remove items from the queue by identifier

grahamalama:


alexcottner (approved):

leplatrem (approved):
We could add this new endpoint to the test `test_get_protected_endpoints()`

grahamalama:
@leplatrem good call, though instead of adding a case to that test, I added a new `test_delete_queue_item_by_id_requires_authn`.

PR #1048, opened 2024-09-14
Include new and existing ticket steps

cbellini:
it looks like we might also need to include steps for new and existing tickets to map statuses

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -437,6 +437,21 @@
   parameters:
     jira_project_key: SNG
     labels_brackets: both
+    status_map:
+      UNCONFIRMED: Backlog
+      NEW: To Do
+      ASSIGNED: In Progress
+      REOPENED: In Progress
+      RESOLVED: Done
+      VERIFIED: Done
+      FIXED: Done
+      INVALID: Cancelled
+      WONTFIX: Cancelled
+      INACTIVE: Cancelled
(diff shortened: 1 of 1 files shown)


leplatrem (approved):

leplatrem:
@cbellini  your branch seems to be a bit outdated. Can you have a look please?

leplatrem:
@cbellini the synchronization now fails with `HTTP 400: Field 'customfield_10716' cannot be set. It is not on the appropriate screen, or unknown.`

Apparently that there could be two `Severity<space>` fields. One with a trailing space, (10319) aka `Severity ` and one without (10716).
Shall we reconfigure JBI to use the other field? `10319`? Or can you reconfigure your Jira project?

cbellini:
> Apparently that there could be two `Severity<space>` fields. One with a trailing space, (10319) aka `Severity ` and one without (10716). Shall we reconfigure JBI to use the other field? `10319`? Or can you reconfigure your Jira project?

@leplatrem I'll determine which severity field we'll use and update here shortly.

leplatrem:
I talked to a Jira yesterday and it's very likely to be `customfield_10319`. I'll deploy the change and see how it goes. The webhooks event won't be lost, we have a retry queue.

cbellini:
Sounds good. I'll keep SDD-25963 open for the time being, but I like the idea of trying 10319 out.

PR #1075, opened 2024-12-04
Fix retry job runtime error

leplatrem:


.github/workflows/ci.yaml
```diff
@@ -40,6 +40,8 @@ jobs:
         run: make install
       - name: Run tests
         run: bin/test.sh
+      - name: Run retry
+        run: DL_QUEUE_CONSTANT_RETRY=false poetry run python -m jbi.retry.py
```
grahamalama, line 44: Don't know how much this matters in the grand scheme of things, but would it be better to explicitly use the Python in our .venv, like we do in the docker container? Don't know if `poetry run` does any side effect magic 🤔.

grahamalama (commented):
Looks like we only need to provide some missing env variables for the script to run, then we're good to go

grahamalama (approved):

alexcottner (approved):

PR #1076, opened 2024-12-04
Publish to GAR along with Dockerhub

grahamalama:
This will also require setting a `GCPV2_GITHUB_WORKLOAD_IDENTITY_PROVIDER` [repository variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#creating-configuration-variables-for-a-repository).

### DIFF
--- a/.github/workflows/build-publish.yaml
+++ b/.github/workflows/build-publish.yaml
@@ -16,13 +16,20 @@ jobs:
     env:
       TEST_TAG: ${{ github.repository }}:test
       TEST_CONTAINER_NAME: jbi-healthcheck
+      GAR_LOCATION: us
+      GAR_REPOSITORY: jbi-prod
+      GCP_PROJECT_ID: moz-fx-jbi-prod
+      IMAGE: jbi
     steps:
       - name: Check out the repo
         uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Fetch everything (tags)
+          fetch-tags: true
+      
       - name: Set tag version
         run: echo "JBI_TAG=$(git describe --tags --abbrev=4)" >> $GITHUB_ENV
+      
       - name: Build `version.json` file
         run: |
           printf '{\n    "commit": "%s",\n    "version": "%s",\n    "source": "%s",\n    "build": "%s"\n}\n' \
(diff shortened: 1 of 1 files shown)


.github/workflows/build-publish.yaml
```diff
@@ -35,27 +42,43 @@ jobs:
...
             type=sha,prefix=,enable={{is_default_branch}}
 
+      - id: gcp_auth
+        name: Log into GCP
+        uses: google-github-actions/auth@v2
+        if: github.event_name != 'pull_request'
+        with:
+          token_format: access_token
+          service_account: artifact-writer@${{ env.GCP_PROJECT_ID }}.iam.gserviceaccount.com
+          workload_identity_provider: ${{ vars.GCPV2_GITHUB_WORKLOAD_IDENTITY_PROVIDER }}
```
leplatrem, line 61: This looks a bit different than:

https://github.com/mozilla-services/telescope/blob/ba8b8ac7914a039db41c8ca3436dbba3d65f2ec7/.github/workflows/build-publish.yaml#L63-L78

Is this the preferred way?
Should we align both?

.github/workflows/build-publish.yaml
```diff
@@ -35,27 +42,43 @@ jobs:
...
         with:
-          images: ${{ github.repository }}
+          images: 
+            ${{ github.repository }}
+            ${{ env.GAR_LOCATION }}-docker.pkg.dev/${{ env.GCP_PROJECT_ID }}/${{ env.GAR_REPOSITORY }}/${{ env.IMAGE }}
           # https://github.com/marketplace/actions/docker-metadata-action#tags-input
           tags: |
             type=semver,pattern={{raw}}
             type=raw,value=latest,enable={{is_default_branch}}
             type=sha,prefix=,enable={{is_default_branch}}
```
leplatrem, line 52: Don't we set tag here? https://github.com/mozilla-services/telescope/blob/ba8b8ac7914a039db41c8ca3436dbba3d65f2ec7/.github/workflows/build-publish.yaml#L56C18-L58C29

leplatrem (approved):
Some differences with the way we do it in Telescope. Happy to align it with this one if that's a more concise way to accomplish the same

.github/workflows/build-publish.yaml
grahamalama: This tagging strategy matches what we do with Dockerhub, since I only added another image repository. Happy to add the `JBI_TAG` tag here to keep things consistent with other syseng projects, though

grahamalama: I'm going to merge this as-is, and if we want to revert back to `git describe --tags --abbrev=4` tags, we can do so 👍🏻

PR #1098, opened 2025-01-28
Update components and issue status for FOG Migration

Dexterp37:
The component was not getting synched, so I hope this fixes it :)

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -685,9 +685,11 @@
       new:
         - create_issue
         - maybe_delete_duplicate
+        - maybe_update_components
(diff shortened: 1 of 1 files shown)


config/config.prod.yaml
```diff
@@ -685,9 +685,11 @@
       new:
         - create_issue
         - maybe_delete_duplicate
+        - maybe_update_components
```
leplatrem, line 688: Don't you want this for existing bugs that get modified too?

leplatrem (approved):
Sorry, I didn't catch that `maybe_update_components` was missing in #1095

PR #1134, opened 2025-04-28
Making the Jira character limit configurable and reducing for addons team per request

alexcottner:
Making the jira character limit configurable and reducing to 10k for the addons team as requested.
[Link](https://mozilla.slack.com/archives/C016BTJKM9B/p1745869073069289?thread_ts=1745858899.387559&cid=C016BTJKM9B) to thread for context.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -6,6 +6,7 @@
   parameters:
     jira_project_key: WEBEXT
     labels_brackets: both
+    jira_char_limit: 10000
 
 - whiteboard_tag: fidedi
   bugzilla_user_id: tbd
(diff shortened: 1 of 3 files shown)


Rob--W:
For context: when I added `[addons-jira]` to https://bugzilla.mozilla.org/show_bug.cgi?id=1942712, jbi failed to create a JIRA ticket, with error `POST /rest/api/2/issue?updateHistory=false -> HTTP 400: The entered text is too long. It exceeds the allowed limit of 32,767 characters.`

https://github.com/mozilla/jira-bugzilla-integration/pull/1132 tried to fix that by slightly lowering the limit, but that still failed.

So I wondered, why don't we just cut down the amount of text copied to say 10000 (or 15000), since Bugzilla is still there as the single source of truth, and that's how we arrived at this PR.

leplatrem (approved):

PR #1142, opened 2025-05-22
Adding type map for hnt, which only has tasks in Jira

alexcottner:
Received this error for an HNT bug this morning. It looks like that project only has Task types in Jira.

```
{
  errorMessages: [],
  errors: {
    issuetype: "The issue type selected is invalid."
  }
}
```


leplatrem (approved):
Good call 👌

PR #1149, opened 2025-06-10
Add uniffi to config.prod.yaml

ddurst:
Before a reorg, the team working on UniFFI was having jbi sync things to the DISCO Jira project via the disco whiteboard tag.

After that reorg, we would like to differentiate between things that sync to DISCO and things that should be UniFFI-specific. Those things should be able to sync to the UNIFFI Jira project.

This patch adds support for a `uniffi` whiteboard entry to allow jbi to sync things to UNIFFI.


leplatrem (approved):
Thank you for opening the PR 🙏

Issue #1157, opened 2025-06-30
Resolution is not set in Jira when bugzilla tickets are closed


muffinresearch:
I think this has been raised in slack before, but since I don't see a ticket for this issue [(aside from a mention here](https://github.com/mozilla/jira-bugzilla-integration/issues/200#issuecomment-1225956440)) I'm filing this new issue.

On some projects there's an issue where the resolution is not being set when Bugzilla tickets are closed. I think JBI is handling closing Jira tickets when bugzilla issues are closed rather than something solely on the JIRA side, but it would be good to understand what's happening and where the fix is needed. It's possible this could be a Jira configuration issue instead of JBI.

To illustrate the issue compare these two JQL queries:

- [RECOMP Issues closed with no resolution - should be empty](https://mozilla-hub.atlassian.net/issues/?jql=project%20%3D%20RECOMP%20and%20status%20%3D%20Closed%20and%20Resolution%20%3D%20null)
- [FIDEFE Issues closed with no resolution](https://mozilla-hub.atlassian.net/issues/?jql=project%20%253D%20FIDEFE%20and%20status%20%253D%20Closed%20and%20Resolution%20%253D%20null)

For some reason FIDEFE seems to have issues closed but the resolution is not set correctly, RECOMP which was setup later on, doesn't have this problem, not sure if something is fixing this on on the 

I don't see anything automation related on either of these two specific projects.

Here's their respective JBI configs:

* https://github.com/mozilla/jira-bugzilla-integration/blob/619406629d9886d72cc1b990cb0c4cabe4a226b9/config/config.prod.yaml#L18-L52
* https://github.com/mozilla/jira-bugzilla-integration/blob/619406629d9886d72cc1b990cb0c4cabe4a226b9/config/config.prod.yaml#L384-L412

Note: this problem is much broader than just FIDEFE too: https://mozilla-hub.atlassian.net/issues/?jql=status%20%3D%20Closed%20AND%20Resolution%20%3D%20EMPTY%20ORDER%20BY%20updated%20DESC

The impact of this issue is that reports like `created vs resolved` don't work if tickets are not resolved.

alexcottner:
[...]

PR #1165, opened 2025-07-22
Update FIDEFE config to set resolutions

muffinresearch:
See also #1157 - if this update works this should resolve that problem for FIDEFE. 

Before this is landed, from experiences with other projects that have enabled this, is there anything that I should look out for in terms of failures when this is enabled?

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -28,6 +28,7 @@
         - add_link_to_jira
         - maybe_assign_jira_user
         - maybe_update_issue_status
+        - maybe_update_issue_resolution
         - maybe_update_issue_priority
         - maybe_update_issue_severity
         - maybe_update_issue_points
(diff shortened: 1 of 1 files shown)


config/config.prod.yaml
```diff
@@ -50,6 +52,7 @@
       INCOMPLETE: Closed
       WORKSFORME: Closed
       REOPENED: In Progress
+    resolution_map: *basic-resolution-map
```
smarnach, line 55: This references the anchor before it's defined. You need to move the definition here.

leplatrem:
:red_circle: [HTTPError](https://mozilla.sentry.io/issues/6777898796/?referrer=slack&notification_uuid=bdfc2dd2-7266-4631-9e39-60035376d9b1&environment=prod&alert_rule_id=11827252&alert_type=issue)
jbi.jira.client in raise_for_status
PUT /rest/api/2/issue/FIDEFE-5606?notifyUsers=true -> HTTP 400: Field 'resolution' cannot be set. It is not on the appropriate screen, or unknown.

@muffinresearch are you sure the resolution field is on create and update form?

alexcottner:
Reverting this commit as we had over 30 events unable to be processed.

PR #1166, opened 2025-07-23
Update FXVPN2 config 

lesleyjanenorton:
This PR updates the config for syncing bugs to the FXVPN2 Jira project.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -732,5 +732,37 @@
   description: VPN Integration in Firefox
   parameters:
     jira_project_key: FXVPN2
+    steps:
+      new:
+        - create_issue
+        - maybe_delete_duplicate
+        - add_link_to_bugzilla
+        - add_link_to_jira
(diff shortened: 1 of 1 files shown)


config/config.prod.yaml
```diff
@@ -732,5 +732,37 @@
...
-    status_map: *basic-status-map
+    status_map:
+      UNCONFIRMED: Backlog
+      NEW: Backlog
+      ASSIGNED: In Progress
+      REOPENED: In Progress
+      RESOLVED: Done
+      VERIFIED: Done
+      FIXED: Done
+      INVALID: Cancelled
```
alexcottner, line 762: It doesn't look like you have a cancelled status in Jira (yet?).
image

leplatrem: `Cancelled` in not listed in the *Create* form, which sort of makes sense. We should be good, I don't think a bugzilla ticket will be created as `Invalid`.

PR #1167, opened 2025-07-23
Ref #984: Show more details about attachment when created 

leplatrem:
Ref #984


muffinresearch:
@leplatrem with this released should I expect to see any changes with how attachment comments work? Noting that attachment modifications are still dependent on https://bugzilla.mozilla.org/show_bug.cgi?id=1979028

PR #1207, opened 2025-10-14
Update priority and severity for PCF project

davehunt:
This is an attempt to update priority and severity for the PerfCompare (PCF) project. I'm waiting on a couple of service desk requests before this can be merged, and I have a question about the default priority map (I'll add this to the relevant line in the PR).

- [x] Add severity custom field to PerfComare (PCF) issue forms ([SDD-37128](https://mozilla-hub.atlassian.net/servicedesk/customer/portal/4/SDD-37128)).
- [x] Add P5 option for Priority field ([SDD-37129](https://mozilla-hub.atlassian.net/servicedesk/customer/portal/4/SDD-37129)).

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -364,6 +364,8 @@
         - maybe_assign_jira_user
         - maybe_update_issue_status
         - maybe_update_issue_resolution
+        - maybe_update_issue_priority
+        - maybe_update_issue_severity
         - sync_whiteboard_labels
         - sync_keywords_labels
       existing:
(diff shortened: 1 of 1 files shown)


leplatrem (approved):
I also assume that P4 and P5 didn't exist when jbi was developped.

If we are sure that the P4 and P5 options will be available accross all projects consistently, I agree that we could switch the default mapping.

But let's play it safe, let's configure it for your project only. Roll out. Double check, and then switch all projects in a different PR

PR #1215, opened 2025-10-23
Adding bz-ffx-core and bz-ffx-firefox actions for syncing Core and Firefox bugs to the BZFFX project

dklawren:


dklawren:
The run_lint test shows no errors but yet it exited as non-zero so not sure what is failing with my change.

PR #1228, opened 2025-11-18
fix: Jira.create_component called with project_key and should be just project

dklawren:
The param `project_key` should be `project`. Misread the docs somewhere else and instead should have consulted https://jira.readthedocs.io/api.html#jira.client.JIRA.create_component. Of course didn't catch this until the change was used by my config steps.


leplatrem:
> Of course didn't catch this until the change was used by my config steps.

Classic JBI development situation 🙃

leplatrem (approved):

PR #1242, opened 2025-12-17
fix: Do not return see_also values not in the same project as whiteboard tag

dklawren:
This patch changes the current behavior of returning all see_also values from a bug regardless of jira project. Now if a whiteboard tag matches a set of actions in the config, and a bug is returned with a se_also value for say BZFFX and the whiteboard tag is for GENAI, the BZFFX see_also will not be returned and the actions will create a new issue in the GENAI project. Before it would skip it because the BZFFX one already existed but it not related at all. 

Let me know if you think there is a possibly better solution for this. Thanks


dklawren:
> This will cause duplicated Jira tickets to be created. For example. if a whiteboard tag for project1 is replaced with project2.

I supposed I could a special case in for BZFFX but it feels bad to do so. If someone has a different idea I would be willing work implement it.

alexcottner:
> > This will cause duplicated Jira tickets to be created. For example. if a whiteboard tag for project1 is replaced with project2.
> 
> I supposed I could a special case in for BZFFX but it feels bad to do so. If someone has a different idea I would be willing work implement it.

I don't have any better ideas unless the BMO plugin behavior can change. We can merge this and release next week to see how it goes.

PR #1247, opened 2026-01-05
Add steps to mirror Bugzilla dependencies in Jira when possible

davehunt:
This PR implements automatic creation of Jira "Blocks" issue links based on Bugzilla's `depends_on` and `blocks` fields. This is an opt-in feature that syncs dependency relationships during both CREATE and UPDATE operations.

Currently, when bugs have dependency relationships in Bugzilla (via depends_on and blocks fields), these relationships are not reflected in the linked Jira issues. This makes it difficult to track dependencies and blockers in Jira, requiring manual link creation or constant cross-referencing with Bugzilla.

This initial implementation only adds links. When dependencies are removed from Bugzilla, corresponding Jira links remain. If the webhook contains details of removed dependencies then it should be possible to build on this to remove links in a future enhancement.


JCMOSCON1976 (approved):
Approved

PR #1252, opened 2026-01-14
Enable component creation for FXP and FXPE projects

davehunt:
This change adds `create_components: true` to the `jira_components` configuration for both the FXP (Performance Team) and FXPE (Performance Engineering Team) projects.

When enabled, the `maybe_update_components` step will automatically create components in Jira if they don't already exist, rather than silently skipping missing components.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -154,6 +154,7 @@
     jira_components:
       use_bug_component: false
       use_bug_component_with_product_prefix: true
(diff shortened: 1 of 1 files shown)


JCMOSCON1976 (approved):
LGTM.
Please make sure that the JBI service account has permission to create components in your Jira project.

JCMOSCON1976 (approved):
LGTM

PR #1263, opened 2026-01-28
Update the default priority map now that Jira has options for P4 and P5 values

### LINKED ISSUE #1261: Update default priority map to include P4 & P5 values
There are now values for "P4" and "P5" in Jira, so the default priority map should be updated to use them:

https://github.com/mozilla/jira-bugzilla-integration/blob/4e6305a9abff685c3bfcfa901657973e4d397d37/jbi/models.py#L100-L108

davehunt:
This also removes the priority map override used by the PCF project as it now matches the default. This fixes https://github.com/mozilla/jira-bugzilla-integration/issues/1261 and was previously discussed in https://github.com/mozilla/jira-bugzilla-integration/pull/1207#pullrequestreview-3340544494.

### DIFF
--- a/config/config.prod.yaml
+++ b/config/config.prod.yaml
@@ -434,14 +434,6 @@
         - create_comment
     status_map: *basic-status-map
     resolution_map: *basic-resolution-map
-    priority_map:
-      "": None
-      "--": None
-      P1: P1
-      P2: P2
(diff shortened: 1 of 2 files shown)


JCMOSCON1976 (approved):
LGTM

PR #1264, opened 2026-01-28
Update severity map to support clearing Jira field

### LINKED ISSUE #1262: Update default severity map to allow for None
The current map converts "" and "--" to "N/A", but "N/A" is a valid and distinct value from the severity being not set. I would expect "--" to result in the severity field being cleared.

https://github.com/mozilla/jira-bugzilla-integration/blob/4e6305a9abff685c3bfcfa901657973e4d397d37/jbi/models.py#L110-L117

davehunt:
- Change severity_map to support None values for clearing fields
- Map empty severity values ('', '--') to None by default
- Add 'N/A' mapping to severity_map

This allows users to configure what empty severity values map to, with the default behavior being to clear the Jira severity field. Users can override this by setting a different mapping in their action configuration (e.g., empty string maps to 'N/A' instead of None). This fixes #1262.

@JCMOSCON1976 as this is more than a simple configuration change, I would like to be able to test this end-to-end. Is there a way for me to do this?


JCMOSCON1976 (approved):
LGTM

PR #1267, opened 2026-02-05
#1266 - Add links to github attachments in Jira

boek:
Keeping this a WIP until I figure out a way that I can run the environment locally.

### DIFF
--- a/jbi/bugzilla/models.py
+++ b/jbi/bugzilla/models.py
@@ -103,6 +103,37 @@ class WebhookAttachment(BaseModel, frozen=True):
     is_patch: bool
     is_private: bool
 
+    def is_github_pull_request(self) -> bool:
+        """
+        Returns True if this attachment is a github pull request attachment.
+
+        We identify an attachment as a pull request if the content type contains "github-pull-request"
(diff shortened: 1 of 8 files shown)


jbi/bugzilla/models.py
```diff
@@ -103,6 +103,37 @@ class WebhookAttachment(BaseModel, frozen=True):
...
+                extra={
+                    "bug": {
+                        "id": self.id,
+                    }
+                },
+            )
+            return None
+
+        revision_id = match.group(0)
+        return f"{base_url}/{revision_id}"
```
segunfamisa: I just realized this now, after checking out what the attachment looks like on this bug - https://bugzilla.mozilla.org/show_bug.cgi?id=2014529

It seems that we may need to account for other GitHub projects too. And I think we need to infer the GitHub repository from the `description` of the attachment.


Image

boek: Good point, I didn't consider that this would be used on multiple repositories. Will make an update!

segunfamisa:
@boek  👀

PR #1285, opened 2026-03-25
fix: Correct inward/outward direction for Jira 'Blocks' issue links

davehunt:
Jira's REST API treats inwardIssue as the blocker (the issue that "blocks"), not outwardIssue as the documentation suggests. The original code had these reversed, causing Jira to display the link in the wrong direction (e.g. FXP-4414 blocking FXP-4820 instead of the other way around).

Swap inwardIssue and outwardIssue in create_issue_link_blocks so that the blocking issue is correctly set as inwardIssue, and update tests to match.

### DIFF
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -444,8 +444,8 @@ class JiraService:
             self.client.create_issue_link(
                 data={
                     "type": {"name": "Blocks"},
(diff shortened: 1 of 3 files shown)


JCMOSCON1976 (approved):
Approved

PR #1287, opened 2026-03-27
Remove Jira 'Blocks' links when depends_on/blocks removed in Bugzilla

davehunt:
When a bug is removed from depends_on or blocks in Bugzilla, delete the corresponding 'Blocks' link in Jira. Previously only link creation was handled.

Also fixes incorrect inward/outward assertion in test_sync_depends_on_links_handles_multiple_jira_issues.


JCMOSCON1976 (approved):
LGTM

PR #1296, opened 2026-04-13
Rename genai/GENAI to aife/AIFE

Mardak:
not sure what's the best timing/sequencing but we should be able to rename the project just after this merges


JCMOSCON1976 (approved):
LGTM

PR #1297, opened 2026-04-20
Fix CI test target and update tests for FastAPI strict_content_type

davehunt:
- Fix CI workflow typo: `make tests` → `make test`. The test job was silently succeeding without running any tests.
- Update 9 tests across `test_router.py`, `test_app.py`, and `test_retry.py` to use `content=<json>` with `Content-Type: application/json` instead of `data=<json_string>`. FastAPI 0.132.0 (shipped in the recent 0.120.2 → 0.135.3 bump) introduced strict content-type checking that rejects JSON bodies sent without the header, causing all 9 tests to fail.


JCMOSCON1976 (approved):
LGTM

PR #1307, opened 2026-04-27
Fix duplicate Relates links by normalising key order

davehunt:
- Bugzilla mirrors `see_also` links bidirectionally: adding Bug B to Bug A's `see_also` automatically adds Bug A to Bug B's `see_also`, firing two webhook events
- Both events tried to create a 'Relates' link — `(FXP-1, FXP-2)` and `(FXP-2, FXP-1)` — which Jira stores as two separate links
- Fix: sort the issue keys before calling the Jira API so both events produce the identical request; Jira returns 400 on the second (duplicate), which is already handled

### DIFF
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -560,12 +560,17 @@ class JiraService:
     ):
         """Create a 'Relates' link between two Jira issues.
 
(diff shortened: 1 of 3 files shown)


JCMOSCON1976 (approved):
LGTM

PR #1308, opened 2026-04-27
Enable sync_see_also for FXPE, FP, and PCF projects

davehunt:
- Adds `sync_see_also` to `new` and `existing` steps for FXPE, FP, and PCF projects


JCMOSCON1976 (approved):
LGTM

PR #1330, opened 2026-05-27
Update status_map for FXDROID to match recent changes

segunfamisa:
We recently had some changes in the jira mapping, and it required changing the `status_map` for FXDROID board


JCMOSCON1976 (approved):
LGTM

PR #1331, opened 2026-05-27
feat: Sync Bugzilla regression relationships to Jira Problem/Incident links

davehunt:
Adds `sync_regressions` step that creates/removes Jira "Problem/Incident" links based on Bugzilla's `regressions` and `regressed_by` fields. Handles both CREATE and UPDATE events since regression links can be set when filing a new bug.


JCMOSCON1976 (approved):
LGTM

PR #1337, opened 2026-06-03
fix: Query outward issue when deleting Duplicate and Problem/Incident links

davehunt:
Jira returns only the OTHER issue in each link object when fetching an issue's links. Both `delete_issue_link_duplicates` and `delete_issue_link_causes` were querying from the inward side, so Jira returned only `outwardIssue` — the `inwardIssue` key check then never matched. Fix: query from the outward side (original_issue / caused_issue) to get `nwardIssue` pointing to the issue to match. Also corrects test mocks which incorrectly included both sides.

See https://developer.atlassian.com/cloud/jira/platform/issue-linking-model/ for the issue linking model, which shows only one of `inwardIssue` or `outwardIssue` is present on the API response.

### DIFF
--- a/jbi/jira/service.py
+++ b/jbi/jira/service.py
@@ -622,18 +622,19 @@ class JiraService:
     def delete_issue_link_duplicates(
         self, context: ActionContext, duplicate_issue: str, original_issue: str
     ):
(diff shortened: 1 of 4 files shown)


JCMOSCON1976 (approved):
LGTM

PR #1344, opened 2026-06-17
feat: Combine sync_depends_on_links and sync_blocks_links into sync_dependencies

davehunt:
Replaces two separate dependency syncing steps with a single `sync_dependencies` step, consistent with how `sync_regressions` and `sync_duplicates` handle their paired fields.

Updates all projects in prod config that used the old steps: FIDEFE, FXP, FXPE, FP, PCF, RECOMP, and the GenAI group (AIFE and others sharing the genai-steps anchor).


JCMOSCON1976 (approved):
Approved

PR #1346, opened 2026-06-17
Update nimbus configuration; add nimbus-sdk tag

freshstrangemusic:
This patch also updates the bugzilla user ID for the nimbus project because (a) the formerly listed user no longer works here and (b) I am the triage owner of both of these components.


JCMOSCON1976 (approved):
LGTM

PR #1349, opened 2026-06-19
fix: Fully sync fields when issue is created via a modify webhook

davehunt:
## Problem

When a whiteboard tag is added to a pre-existing Bugzilla bug, the Jira issue is created in response to a bug-modify webhook (not bug-create). The runner correctly treats this as a `CREATE`, but `event.changes` is still populated with the whiteboard change.

Several `sync_*` steps decided between full-sync (create) and delta-sync (update) by testing the presence of webhook changes (`if context.event.changes:`) rather than `context.operation`. In the create-via-modify case `event.changes` is truthy, so they took the delta path, found no change to their own field, and returned `NOOP` — dropping keywords (e.g. `regression`), pre-existing whiteboard tags, and links.

## Fix

Gate the delta path on `context.event.changes and context.operation == Operation.UPDATE` so a CREATE always syncs the full field value. Applied to `sync_keywords_labels`, `sync_whiteboard_labels`, `sync_see_also`, `sync_regressions`, `sync_depends_on_links`, `sync_blocks_links`. `sync_duplicates` is unchanged (dupe fields are never set on creation).

## Tests

Added regression tests for the create-via-modify scenario, and updated existing `sync_see_also` tests to set `operation=Operation.UPDATE`. All 386 tests pass (94% coverage); ruff and mypy clean.


JCMOSCON1976 (approved):
Approved

PR #1361, opened 2026-06-29
fix: Fully sync fields when whiteboard tag is added to a bug with an existing Jira issue

davehunt:
- When a whiteboard tag is added to a bug that already has a linked Jira issue in `see_also`, the runner previously ran `existing` (UPDATE) steps, which only react to the fields changed in that single webhook event — leaving all other fields stale.
- This covers two cases: a tag that was removed and later restored, and a tag added simultaneously with a `see_also` link to connect an existing bug to an existing Jira issue.
- Adds `_tag_added_to_whiteboard()` to detect when the whiteboard change transitions the project tag from absent to present, and sets `operation=CREATE` so all `new` steps run and sync all current bug field values unconditionally.
- Modifies `create_issue` to handle an already-linked issue: updates the summary and returns early rather than creating a duplicate. All other `new` steps then resync their respective fields via their normal CREATE-mode logic.


JCMOSCON1976 (approved):
Approved

PR #1363, opened 2026-06-29
feat: Sync Bugzilla issue type changes to existing Jira issues

davehunt:
`issue_type_map` was only consulted when **creating** a Jira issue (the `create_issue` step). Once an issue existed, a later change to the Bugzilla bug's `type` field never propagated to the linked Jira issue.

This adds a new **opt-in** step, `maybe_update_issue_type`, that on `UPDATE` events — when the `type` field actually changed — maps `bug.type` through `issue_type_map` (falling back to `"Task"`, mirroring `create_issue`) and updates the Jira issue type via the existing `update_issue_field` helper.

The step is NOOP on create (creation still owns the initial type) and NOOP when `type` didn't change. It is **not** wired into the defaults or any action config — teams opt in by adding it to their `existing` steps.


JCMOSCON1976 (approved):
Approved

PR #1364, opened 2026-06-29
fix: Pass global_id to make Bugzilla remote link idempotent

davehunt:
- `add_link_to_bugzilla` called `create_or_update_issue_remote_links` without a `global_id`, causing each invocation to create a new remote link on the Jira issue rather than upserting the existing one.
- Pass `global_id=str(bug.id)` so the Jira API treats repeated calls (e.g. during a full field resync triggered by a whiteboard tag add) as updates to the same link.

**Note on existing data:** Links created before this fix have no `global_id`. The first call after this fix will create a second link for those issues; subsequent calls will then be idempotent. This one-time duplication was deemed acceptable.


JCMOSCON1976 (approved):
Approved

PR #1371, opened 2026-07-14
Sync issue type changes to existing issues for performance projects

davehunt:
Add `maybe_update_issue_type` to the existing steps for FXP, FXPE, FP, and PCF so that Bugzilla bug type changes are reflected on the linked Jira issue, using the `issue_type_map` already configured for these projects.


JCMOSCON1976 (approved):
Approved

Issue #1376, opened 2026-08-06
Failed events silently block a bug for hours: slow retries, no visibility, and re-triggering is a no-op


msujaws:
## Summary

When an event fails and lands in the dead letter queue, the affected bug goes completely dark for hours: subsequent events for that bug are never processed live, the reporter has no way to see that anything is queued, and the obvious manual workaround (re-adding the whiteboard tag) is a no-op by design. Three bugs hit this today across three different teams' projects.

## What we observed

On 2026-08-06, three bugs had a valid whiteboard tag added and received no Jira issue for between 48 minutes and 2.5 hours. All three were then created in a single batch between 16:05Z and 16:11Z:

| Bug | Tag added | Jira issue created / linked | Delay |
| --- | --- | --- | --- |
| [2061210](https://bugzilla.mozilla.org/show_bug.cgi?id=2061210) (`[fxdroid]`) | 13:40:36Z | FXDROID-9770 @ 16:09:17Z | 2h29m |
| [2061250](https://bugzilla.mozilla.org/show_bug.cgi?id=2061250) (`[hnt]`) | 14:38:16Z | HNT-2903 @ 16:11:15Z | 1h33m |
| [2060382](https://bugzilla.mozilla.org/show_bug.cgi?id=2060382) (`[fidedi-ope]`) | 15:17:04Z | FIDEDI-1880 @ 16:05:40Z | 48m |

This was not a service outage. Unrelated bugs were processed live within seconds throughout the same window, for example bug 2061252 linked to AIDISC-176 at 14:19:14Z (8 seconds after its tag was added) and bug 2061279 linked to FXDROID-9769 at 14:58:24Z. So some events were failing while others in the same period succeeded.

**Request:** could someone check Sentry for the exceptions on these three queue items between 13:40Z and 16:05Z on 2026-08-06? The underlying failure cause is only visible to JBI operators, and it matters as much as the latency below. Everything that follows is about how the recovery path behaves once an event has failed, independent of why it failed.

## Why this is painful from the reporter's side

For bug 2061250, the sequence was:

1. 14:38:16Z: `[hnt]` added to the whiteboard. Nothing happens.
2. 16:04:18Z / 16:04:33Z: reporter removes and re-adds `[hnt]`, on the assumption the event was dropped. Still nothing happens.
3. 16:11:15Z: the retry job drains the queue and HNT-2903 is created and linked.

[...]

PR #1377, opened 2026-08-06
fix: Reprocess a bug's queued events when a new event arrives

msujaws:
Addresses the fourth point of #1376. Filed after three bugs went unlinked for between 48 minutes and 2.5 hours on 2026-08-06 while unrelated bugs were being processed live in the same window.

## Problem

`DeadLetterQueue.is_blocked` returns true when *any* item exists for a bug id, so a single failed event stops every subsequent event for that bug from being processed until the retry job runs. From the reporter's side the bug simply goes quiet, and the obvious workaround -- removing and re-adding the whiteboard tag -- cannot work: the new event is postponed as well. Bug 2061250 accumulated three queued events (the original tag add, plus a remove and re-add) for one desired outcome, and was only resolved when the retry job drained the queue an hour and a half later.

## Change

When a webhook arrives for a blocked bug, try to drain that bug's pending items before postponing:

```python
if await queue.is_blocked(request):
    if not await drain_bug_queue(queue, request.bug.id, actions):
        await queue.postpone(request, rid=request_id)
        ...
        return {"status": "skipped"}
```

`drain_bug_queue` walks the bug's items oldest first through a new `DeadLetterQueue.retrieve_for_bug()` (a passthrough to `QueueBackend.get()`, which already exists), executes each one, and marks it done. It bails out on the first failure so event ordering is preserved. When the drain fails, behaviour is byte for byte what it is today.

The effect is that touching a stuck bug becomes a real retry trigger, so recovery no longer has to wait for the next scheduled retry pass.

This deliberately leaves the other three points of #1376 alone -- retry cadence, retryable versus terminal error classification, and queue visibility are all larger discussions.

## Tradeoffs worth a reviewer's attention

- The drain runs inline in the webhook handler, so a request for a blocked bug is slower and can hit a still-failing dependency. It is bounded by that one bug's pending items, normally one or two.
- A pod and the retry job can now drain the same bug concurrently. `maybe_delete_duplicate` and the `refresh_bug_data` inside the action are the existing defenses; this widens that window somewhat.
- There is no cap on how many items a single request will drain. Happy to add one if you would prefer a bounded version.

## Tests

- `test_execute_or_queue_blocked` updated: the pending item still fails, so the request is postponed as before and nothing is marked done.
- `test_execute_or_queue_blocked_drains_pending_items` added: the pending item succeeds, so it is marked done, the incoming event is processed, and nothing is postponed.

`ruff check`, `ruff format` and `mypy` are clean. Ten tests in `tests/unit/test_runner.py` fail in my sandbox both with and without this change, because they need network access; everything else passes.

## Docs

Added a paragraph to `docs/troubleshooting.md` under "Bugzilla tickets are not showing up as issues on Jira" describing the queued-event behaviour, since that is where a consumer looks first when a ticket has no Jira link.


JCMOSCON1976:
Thanks for the PR. I will be reviewing it next week.

PR #1378, opened 2026-08-06
Send a custom User-Agent on Bugzilla and Jira requests

JCMOSCON1976:
## Summary
- Sends a custom `User-Agent: jira-bmo-integration/1.0 (+https://github.com/mozilla/jira-bugzilla-integration)` header on every outgoing Bugzilla and Jira API request.
- Lets Bugzilla/Jira operators identify JBI's traffic in their access logs (came up as an alternative to allowlisting JBI's egress IP range, which is shared with other services and not stable).

## Changes
- Added `jbi/common/net.py` with a `USER_AGENT` constant.
- `BugzillaClient.__init__` now sets it on its `requests.Session`.
- `jira.service.get_service()` now builds its `JiraClient` with a pre-configured `requests.Session` carrying the header (the underlying `atlassian-python-api` client only adds auth headers on top of a passed-in session, so the custom UA survives).

## Test plan
- [x] `uv run pytest tests/unit/bugzilla/test_client.py tests/unit/jira/test_client.py tests/unit/jira/test_service.py` — all passing
- [x] Manually verified both `BugzillaClient` and `jira.service.get_service()` set the header as expected
- [x] `ruff check`/`mypy` pass on the changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jbi/bugzilla/client.py
+++ b/jbi/bugzilla/client.py
@@ -4,6 +4,7 @@ import requests
 
 from jbi import environment
 from jbi.common.instrument import instrument
+from jbi.common.net import USER_AGENT
 
 from .models import (
     ApiResponse,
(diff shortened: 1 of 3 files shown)


spatilmoz (approved):
looks good

PR #1386, opened 2026-08-25
docs/bmo jira bidirectional plan

jgauf:
- **docs: add BMO<->Jira bidirectional integration engineering plan**
- **docs: expand plan for review clarity, add metabug/epic semantics**

### DIFF
--- a/docs/bmo-jira-bidirectional-integration-plan.md
+++ b/docs/bmo-jira-bidirectional-integration-plan.md
@@ -1,125 +1,260 @@
 # BMO ↔ Jira Bidirectional Integration — Engineering Plan
 
-> **Status:** DRAFT for review · **Scope:** Expands JBI from one-directional
-> (Bugzilla → Jira) sync into bidirectional, hierarchy-aware sync per the
-> "BMO – Jira Integration" PRD (v2). This document is the engineering plan and
-> traceability map; it does not itself change behavior.
+> **Status:** DRAFT for engineering review.
+> **Purpose:** Translate the "BMO – Jira Integration" PRD (v2) into a concrete,
+> reviewable engineering plan: what exists in JBI today, what the requirements
+> ask for, where the two diverge, and a phased, testable path to close the gap.
+> **Audience:** Reviewers and implementers. This document does not itself change
+> behavior; it is the design and traceability reference the implementation PRs
+> will point back to.
 
 ---
 
 ## 1. Decisions locked
 
-| # | Decision |
+These decisions were settled with stakeholders and constrain everything below.
+They are stated with their reasoning so reviewers can challenge the *reasoning*
+rather than infer it.
+
+| Area | Decision |
 |---|---|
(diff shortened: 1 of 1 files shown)


JCMOSCON1976 (commented):
Per-project Jira Automation rules won't scale. The plan's §12 setup only covers the pilot project, but this pattern requires manually creating and maintaining a separate Automation rule inside every Jira project that opts in — and prod config already spans 30+ projects (WEBEXT, FXP, JB, AIPLAT, VPN, BZFFX, ...). Onboarding a new project would mean asking that project's Jira admin to hand-configure a rule, with no central visibility into which projects have it, whether it's still correctly configured, etc.

A simpler design: create one centrally-owned Automation rule, and use its scope/JQL filter to select which projects' events get forwarded, instead of duplicating the rule per project. Jira Cloud Premium supports "multiple projects" automation rules scoped by JQL — onboarding a new project then becomes a one-line JQL edit instead of a new rule.

image

JCMOSCON1976 (commented):
The Bugzilla→Jira status map isn't reversible. Every configured action's status_map is many-to-one (e.g. RESOLVED, VERIFIED, FIXED, INVALID, WONTFIX, INACTIVE, DUPLICATE, WORKSFORME, INCOMPLETE, MOVED → Done), so given a Jira status there's no way to recover which specific BMO status/resolution to write back — and this differs per project's custom workflow, so a literal reverse of each status_map would mean maintaining a separate reverse map per action.

One option: instead of reversing each project's named statuses, key off Jira's built-in status category (statusCategory.key/colorName), which every status — regardless of a project's custom workflow — belongs to one of three values: new (blue-gray), indeterminate (yellow), done (green). That gives a single, project-agnostic default map instead of N hand-maintained ones:

Blue/gray (new) → BMO NEW (or REOPENED, if the bug was previously resolved)
Yellow (indeterminate) → BMO ASSIGNED
Green (done) → BMO resolved — but green/"Done" is still ambiguous about which resolution (Fixed, Won't Fix, Duplicate, etc.), so this only solves the status side, not the resolution side; that part still needs an explicit default or a human in the loop.

JCMOSCON1976 (commented):
One more thing — the loop-prevention here only covers one direction. You check whether an inbound Jira event was authored by JBI itself, which stops a Jira change from looping back on itself, but I don't see the same check on the Bugzilla side. 

When the reverse sync writes a field into Bugzilla, that's a real change as far as Bugzilla is concerned, so it fires its normal webhook and that goes straight into our existing forward pipeline, which pushes the same value right back into Jira. At best that's a wasted round trip; at worst, if the value doesn't come back identical (see my status-mapping comment above), we'd actually be corrupting the field instead of just doing extra work. 

It should be a simple fix though — the Bugzilla webhook payload already gives us who made the change (WebhookEvent.user.login), so we can add the same "was this us?" check here that you're already doing on the Jira side.

PR #1387, opened 2026-08-26
Update and expand HNT configuration

fchasen:
Adds field syncing and corrects status and type mapping for the HNT project.


JCMOSCON1976 (approved):
LGTM

PR #1395, opened 2026-09-09
Update cringe config for defect issue type and status mapping

moz-jim:
- Removed unused properties
- Reverted `bug` type to `Task` to match actual Jira usage


JCMOSCON1976 (approved):
LGTM

PR #1396, opened 2026-09-11
fix: Skip Jira status transition when issue already in target status

JCMOSCON1976:
## Summary
- Root-caused a recurring dead-letter-queue error: `POST /rest/api/2/issue/HNT-XXXX/transitions -> HTTP 400: 'transition' identifier must be an integer`
- Happens whenever a Bugzilla bug moves between two resolutions that map to the same Jira status (e.g. `RESOLVED` → `VERIFIED`, both mapped to `Live` for the HNT project). `atlassian-python-api`'s `get_transition_id_to_status_name` only looks at transitions reachable from the *current* status, so when the issue is already at the target status it silently returns `None`, which gets POSTed as `{"transition": {"id": None}}` and rejected by Jira with a confusing 400.
- `JiraService.update_issue_status` now checks the issue's current status via a new `get_issue_status` client method first, and no-ops (logs + returns) if it's already at the target status, avoiding the doomed API call.

## Test plan
- [x] Added `test_update_issue_status_skips_transition_when_already_in_target_status`, verified RED (fails against the real HTTP mock sequence with a `ConnectionError` since the fix code path didn't exist yet) then GREEN after the fix.
- [x] Updated `test_update_issue_status_adds_comment_and_resolution_when_cancelled` to mock the new status-check GET call.
- [x] `uv run pytest tests/unit/jira/test_service.py` — 55 passed
- [x] `uv run ruff check` / `uv run mypy` clean on changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)


spatilmoz (approved):
looks good

